Issue #36, opened 2023-07-31
Create directory for FrontEnd layer


mehalyna:

Krimets:
Created directory for FrontEnd layer

PR #37, opened 2023-07-31
#36 create front end directory

Krymets:
Created directory for FrontEnd layer


FrontEnd/README.md
```diff
@@ -0,0 +1,7 @@
+### Rules:
+1. Create a folder with a name of your task.
```
mehalyna, line 2: Take your point, but it seems better to use folder naming appropriate to the name of the page (or component if needed)

mehalyna (changes requested):

FrontEnd/README.md
Krymets: Fixed as recommended

mehalyna (approved):

PR #60, opened 2023-08-07
fix problem

whooaami:


forum/settings.py
```diff
@@ -12,6 +12,7 @@
 
 from pathlib import Path
 import os
```
Lvyshnevska, line 14: Is module os required in settings? Maybe it is sense to remove it?

whooaami: Fixed

Lvyshnevska (approved):

mehalyna (approved):

Issue #41, opened 2023-08-01
Add User Authorization/Authentification


mehalyna:
Due to #2

PR #76, opened 2023-08-10
#41 add user authorization authentication

Lvyshnevska:
Authentication/authorization and user registration implemented.


authentication/models.py
```diff
@@ -1,3 +1,55 @@
...
+from validation.validate_password import validate_password_long, validate_password_include_symbols
 
-# Create your models here.
+
+class CustomUserManager(BaseUserManager):
+
+    def create_user(self, person_email, password=None, **extra_fields):
+        person_email = self.normalize_email(person_email)
+        extra_fields.setdefault("is_staff", False)
+        extra_fields.setdefault("is_superuser", False)
```
Krymets, line 12: Is it possible to remove setdefault? The values are already set by default in CustomUser model

Lvyshnevska: User create methods improved, setdefault removed.

authentication/models.py
```diff
@@ -1,3 +1,55 @@
...
+    def create_user(self, person_email, password=None, **extra_fields):
+        person_email = self.normalize_email(person_email)
+        user = self.model(
+            person_email=person_email, **extra_fields)
+        user.set_password(password)
+        user.save()
+        return user
+
+    def create_superuser(self, person_email, password, **extra_fields):
+        user = self.create_user(person_email, password, **extra_fields)
```
Shushunya, line 18: Why the email for a superuser isn't normalized while for a regular user it's the opposite?

authentication/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+    def create(self, validated_data: dict):
+        comp_registered = validated_data.get("comp_registered")
+        comp_is_startup = validated_data.get("comp_is_startup")
+        if not comp_registered and not comp_is_startup:
+            raise serializers.ValidationError({"error":
+                                               "Please choose who you represent"})
+        if comp_registered and comp_is_startup:
+            raise serializers.ValidationError({"error":
+                                               "Please choose either registered or startup"})
+        if User.objects.filter(person_email=validated_data["person_email"]):
```
Shushunya, line 58: Email check should be the first one to execute. Now this method creates comp variables and checks their value. This inspection is unnecessary if email is already registered.

Shushunya (changes requested):
Change authentication/serializers.py: UserRegistrationSerializer.create method

authentication/models.py
Lvyshnevska: Create_user method is called to create a superuser. And email is normalized there. There is no need to do it twice.

authentication/serializers.py
Lvyshnevska: Email existence check priority changed.

Shushunya (approved):

authentication/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+            raise serializers.ValidationError(
+                {"error": "Email is already registered"})
+        comp_registered = validated_data.get("comp_registered")
+        comp_is_startup = validated_data.get("comp_is_startup")
+        if not comp_registered and not comp_is_startup:
+            raise serializers.ValidationError({"error":
+                                               "Please choose who you represent"})
+        if comp_registered and comp_is_startup:
+            raise serializers.ValidationError({"error":
+                                               "Please choose either registered or startup"})
```
whooaami, line 60: You have implemented some additional checks related to comp_registered and comp_is_startup. This is good, but comparing these values using if-elif-else instead of sequential conditions might be more efficient.

Krymets (approved):

authentication/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        if not self.user:
+            self.user = User.objects.filter(**params).first()
+            if self.user and not self.user.check_password(password):
+                raise serializers.ValidationError({"error":
+                                                   "Email or password is incorrect"})
+        if self.user and self.user.is_active:
+            return attrs
+        raise serializers.ValidationError({"error":
+                                           "Email or password is incorrect"})
```
whooaami, line 106: You can also change it to if-elif-else here.

authentication/serializers.py
Lvyshnevska: if-if statement replaced with if-elif

authentication/serializers.py
Lvyshnevska: It is the logic of Djoser TokenCreateSerializer. Maybe it's no need to change it here.

authentication/models.py
```diff
@@ -1,3 +1,55 @@
...
+        user.is_superuser = True
+        user.save()
+        return user
+
+
+class CustomUser(AbstractBaseUser, PermissionsMixin):
+    person_email = models.EmailField(max_length=50, unique=True)
+    person_name = models.CharField(max_length=50)
+    person_surname = models.CharField(max_length=50)
+    password = models.CharField(
```
popovycholeg, line 29: Do we store password as is? We should store password hash only.

whooaami (approved):

authentication/models.py
Lvyshnevska: Password field removed from CustomUser model.

mehalyna (approved):

Issue #81, opened 2023-08-10
implement functionality of the saved list


VovaD199:
#62 
add the table which stores the reference between speciate company and saved companies (complicated PK)

PR #111, opened 2023-08-17
Revert 99 #81 saved list

Krymets:
Fixed CompanySavedList


profiles/serializers.py
```diff
@@ -25,3 +25,19 @@ class ProfileSerializer(serializers.ModelSerializer):
     class Meta:
         model = Profile
         fields = "__all__"
+
+
+class CompanySavedListSerializer(serializers.ModelSerializer):
```
Shushunya, line 30: Do we need all of this fields her?

profiles/views.py
```diff
@@ -1,61 +1,46 @@
...
-
+from rest_framework import status
+from rest_framework.generics import ListAPIView, RetrieveUpdateDestroyAPIView
+from rest_framework.response import Response
+from .models import CompanySavedList
+from .serializers import CompanySavedListSerializer
+
+
+class SavedCompaniesList(ListAPIView):
+    """
```
Shushunya, line 9: Redundant new line

profiles/views.py
```diff
@@ -1,61 +1,46 @@
...
+from rest_framework.response import Response
+from .models import SavedCompany
+from .serializers import CompanySavedListSerializer
+
+
+class SavedCompaniesList(ListAPIView):
+    """
+    List of saved companies.
+    """
+    def get(self, request):
```
Shushunya, line 12:
```suggestion
    def list(self, request):
```

Change name so it triggers on GET request.

profiles/views.py
Shushunya, line 1: Change method names so they are triggered when the request comes.
Add queryset, serializer_class and permission_classes attributes to SavedCompanyList. and SavedCompanyDetail. 
Documentaion: https://www.django-rest-framework.org/api-guide/generic-views/

profiles/views.py
```diff
@@ -1,61 +1,46 @@
...
+        if SavedCompany.objects.filter(user=user, company_id=pk).exists():
+            return Response({'error': 'Company already in saved list'}, status=status.HTTP_400_BAD_REQUEST)
+
+        serializer = CompanySavedListSerializer(data={'company': pk, 'user': user.id})
+        if serializer.is_valid():
+            serializer.save()
+            return Response({'Company added': serializer.data})
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def delete(self, request, pk):
```
Shushunya, line 37:
```suggestion
    def destroy(self, request, pk):
```

Rename the method.

profiles/views.py
```diff
@@ -1,61 +1,46 @@
...
+        user = request.user
+
+        try:
```
Shushunya, line 40: This try-except block can be replaced with `get_object_or_404`

profiles/views.py
```diff
@@ -1,61 +1,46 @@
...
+class SavedCompaniesDetails(RetrieveUpdateDestroyAPIView):
+    """
+    Add a company to the saved list.
+    Remove the company from the saved list.
+    """
+    def post(self, request, pk):
+        user = request.user
+
+        # Check if the company is already in the user's saved list
+        if SavedCompany.objects.filter(user=user, company_id=pk).exists():
```
Shushunya, line 28: get_object_or_404 Will do that manual check for you.

Shushunya (changes requested):

profiles/views.py
Shushunya, line 1: I've tried to run your branch locally but I constantly get 500 status.

Shushunya (changes requested):

profiles/views.py
Krymets: > I've tried to run your branch locally but I constantly get 500 status.
Because you are not logged in and you are not passing the token in the request

Krymets: > I've tried to run your branch locally but I constantly get 500 status.

token

Shushunya: Unfortunately no, that's not the problem. If I wasn't authorized, it should have been a 401 or 404 status but I have 500.

Krymets: permission_classes = [IsAuthenticated] added, will now be correct

Shushunya (changes requested):

profiles/views.py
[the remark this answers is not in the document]
Krymets: > Why you chose to override `get` not `list` method?

renamed

profiles/views.py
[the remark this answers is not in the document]
Krymets: > Why you chose to override delete, not `destroy` ?

renamed

VovaD199 (approved):

Lvyshnevska (approved):

PR #113, opened 2023-08-18
migrations added

Krymets:
migrations added


whooaami (approved):

.gitignore
```diff
@@ -7,8 +7,6 @@
 /.env
 /.DS_Store
 __pycache__/
-/authentication/migrations/
-/profiles/migrations/
 /node_modules
 package-lock.json
 package.json
```
ProDG, line 12: Must be included.

Krymets: > Must be included.

Do I understand correctly? They are necessary for the evolution of the database schema, but there is no need to include them in the git because developers can generate them with the makemigrations command?

Issue #112, opened 2023-08-17
Implement the functionality of the Filters on Main Page


mehalyna:
Implement the functionality of the filters on the Main Page.
1) Filter for Header
The default state - Головна
1.1. When the user clicks on the Компанії item, the page displays a list of registered companies.
1.2. When the user clicks on the Стартапи item, the page displays a list of startups.

![image](https://github.com/ita-social-projects/Forum/assets/39273210/fab4c834-87a0-4238-b28b-c4076f93b640)

2) Filter for Menu of Activities
The default state absent
When the user clicks on the Menu item, the page displays a list of registered companies (and startups) with the appropriate activities.

![image](https://github.com/ita-social-projects/Forum/assets/39273210/7529f5ef-ddae-496f-b74f-881d206cb89b)

PR #126, opened 2023-08-21
#112 main page header filters

Shushunya:


profiles/models.py
```diff
@@ -36,12 +36,12 @@ class Profile(models.Model):
 
 
 class Activity(models.Model):
-    activity_id = models.IntegerField(primary_key=True)
+    activity_id = models.AutoField(primary_key=True)
```
ProDG, line 39: In Django you shouldn't declare PK in case you need and autoincremental PK. Probably it's not possible to get rid of this in case if it's already there, just saying for the future.

profiles/tests/test_delete_profile.py
```diff
@@ -77,17 +75,19 @@ def setUp(self) -> None:
...
     def tearDown(self) -> None:
         self.client.logout()
 
-    def test_get_all_profiles_authorized_no_queryparam(self):
-        response = self.client.get(self.base_url)
-        self.assertEqual(200, response.status_code)
-        self.assertEqual(1, len(response.data))
+    def test_get_profile_unauthorized(self):
+        self.client.logout()
+        response = self.client.get("".join([self.base_url, str(self.user1_id)]))
```
ProDG, line 84: I recomment to put explicit URLs in client calls. Using `.format()` method with named placeholders.

So, when looking at the test code, it's super clear, what is the structure of the URL and what the URL actually is.

ProDG: ... code duplication is a much lesser evil here. You can enjoy optimizations in other parts of code, but for unittest I recommend to focus on readability. Make unitests code understandable even by non-technical people (it's possible to some extent).

profiles/tests/test_filter_profiles.py
```diff
@@ -0,0 +1,212 @@
+from rest_framework.test import APITestCase, APIClient
+from authentication.models import CustomUser
+from profiles.models import Profile, Category, Activity
+
+
+def dump(response):
```
ProDG, line 6: I strongly recomment:
1. To extract this helper function into a separate file.
2. To include it (mandatory) with # noqa comment into each unittests file which deals with API.

Shushunya: I removed it eventually to keep the code cleaner (forgot to do it before commit). But I'll add separate file with it then , thanks.

profiles/tests/test_filter_profiles.py
```diff
@@ -0,0 +1,212 @@
...
+        # TODO: add check for types of the returned companied [has to be "startup"]
+
+    def test_get_all_profiles_unauthorized_filter_activity_producer(self):
+        self.client.logout()
+        response = self.client.get(f"{self.base_url}?activity_type=producer")
+        filtered_len = len(response.data)
+        self.assertEqual(200, response.status_code)
+        response = self.client.get(self.base_url)
+        producer_type = [response.data[i]["comp_activity"] for i in range(len(response.data)) if 1 in response.data[i]["comp_activity"]]
+        self.assertEqual(len(producer_type), filtered_len)
```
ProDG, line 185: I recomment to take a closer look at the unittests in the example project.

All the code duplication and other stuff which may look _very_ unoptimized actually has a sense. I don't insist, your unittests also are OK and provide a real coverage. But if you're interested, we can have call where I'll explain the idea behind the approach which you can see in Quoter project.

ProDG (approved):

profiles/tests/test_delete_profile.py
Shushunya: Isn't it better to use f-string then so the url will be even more explicit rather than `.format`?

ddianaoo (approved):

Issue #118, opened 2023-08-18
Update CustomUser model and user registration


Lvyshnevska:
Move profile fields from CustomUser model to Profile model. Update User registration according to changes.

PR #127, opened 2023-08-21
#118 Update custom user model and user registration

Lvyshnevska:


authentication/serializers.py
```diff
@@ -1,66 +1,81 @@
...
             "person_surname",
-            "comp_name",
-            "comp_registered",
-            "comp_is_startup",
+            "company"
         )
 
-    def validate(self, attrs):
+    def validate(self, value):
+        custom_errors = {"password": [],
```
ProDG, line 42:
```
custom_errors = defaultdict(list)
```

So you can:
1. Use constructions as `custom_errors["comp_status"].append(...`
2. .. and then at the end just `if custom_errors: raise ...`

authentication/tests/test_user_login.py
```diff
@@ -0,0 +1,62 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        cls.test_user = CustomUser.objects.create_user(
+            person_email="test@test.com",
+            password="Test1234",
+            person_name="Test",
+            person_surname="Test",
+        )
+        cls.test_user.is_active = True
+        cls.test_user.save()
```
ProDG, line 17: I strongly recommend to not use `setUpTestData()` (and use `setUp`) instead.

Yes, it will make tests execution significantly longer (especially if we create a lot of stuff in setUp). But there is a lot less possible sideeffects.

Let's say, if someone will update or remove instances created in `setUpTestData` — you'll get flaky tests (sometimes they will pass, sometimes not, depending on order of execution).

I use `setUpTestData` only in cases when there are strong reasons for this. In general, tests are pretty fast, so `setUp` is the recommended appoach (as it is a significantly safer one).

authentication/tests/test_user_login.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_login_successful(self):
+        url = "/api/auth/token/login/"
+        data = {
+            "person_email": "test@test.com",
+            "password": "Test1234",
+        }
+        response = self.client.post(url, data)
+        self.assertEqual(response.status_code,
+                         status.HTTP_200_OK)
+        self.assertContains(response, "auth_token")
```
ProDG, line 28: I recommend to have at least one entry in unittests, containing a full JSON returned by BE.

Thus, unittests can be used as a references (for example — a kind of documentation for FE team, or for QC).

In case if there is stuff content of which is hard to predict, you case use `AnyStr`, `AnyInt` etc classes (see example project).

authentication/tests/test_user_login.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_login_email_incorrect(self):
+        url = "/api/auth/token/login/"
+        data = {
+            "person_email": "tost@test.com",
+            "password": "Test1234",
+        }
+        response = self.client.post(url, data)
```
ProDG, line 36: The `url` and `data` variables are redundant.

I recommend to put explicit values directly in calls. It's a bit easier to read what's going on, when the content is in-place instead of being somewhere above in variables.

Also (please share it with ther rest of the team) — from my experience, unittests are the place where it's better to write as explicit code, as possible. Ignoring things as code duplication etc.

I found it very useful to keep unittests in the shape which is more or less readable by people who has no experience in BE, Python or even programming at all (and it _is_ possible to keep it this way).

ProDG (approved):

Shushunya (approved):

authentication/serializers.py
```diff
@@ -1,66 +1,81 @@
...
 from rest_framework import serializers
 
+from profiles.models import Profile
 from validation.validate_password import validate_password_long, validate_password_include_symbols
 
 
 User = get_user_model()
 
 
+class CustomProfileSerializer(serializers.ModelSerializer):
```
Shushunya, line 15: So we have now two separate serializers for Profile? Can you explain a bit why?

Shushunya (approved):

whooaami (approved):

authentication/serializers.py
Lvyshnevska: Fixed. Thanks.

authentication/tests/test_user_login.py
Lvyshnevska: setUpTestData() replaced with setUp()
Thanks!

authentication/tests/test_user_login.py
Lvyshnevska: I've used approach as in example project and added one more assert with full JSON.
Thanks.

authentication/tests/test_user_login.py
Lvyshnevska: Redundant variables were removed from all the tests. 
Thank you for your recommendations a lot.

authentication/serializers.py
Lvyshnevska: The CustomProfileSerializer is used as nested in UserRegistrationSerializer to handle the company fields, that is needed while user registration.

PR #134, opened 2023-08-22
Added tests for the SavedCompany model

Krymets:


profiles/tests/test_models.py
```diff
@@ -0,0 +1,163 @@
...
+        # login users & get tokens
+        self.token1 = self.client.post(
+            path="/api/auth/token/login/",
+            data={
+                "person_email": "testuser1@test.com",
+                "password": "testuser1"
+            }).data["auth_token"]
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token1}")
+        self.user1_id = self.client.get(
+            path="http://127.0.0.1:8000/api/auth/users/me/").data["id"]
```
ProDG, line 77: 1. Use relative URLs `/api/...`.
2. Tokens clearly look as pain in the ass. Again, I recommend to not use them. With ordinary authentication it's as easy as just `self.client.force_login(user_instance)`.
3. I strongly recommend to authenticate users in-place where it's needed. I.e. in each testcase, instead of `setUp` etc.

profiles/tests/test_models.py
```diff
@@ -0,0 +1,163 @@
...
+                "password": "testuser2"
+            }).data["auth_token"]
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token2}")
+        self.user2_id = self.client.get(
+            path="http://127.0.0.1:8000/api/auth/users/me/").data["id"]
+
+
+    def test_add_test_company1_to_saved_list_by_test_user2(self):
+        response = self.client.post(
+            self.url,
```
ProDG, line 92: I recommend to use explicit URLs (write in-place instead of putting into a variable).

ProDG (approved):

BelousSofiya (approved):

RedSSD (approved):

profiles/tests/test_models.py
Krymets: > I recommend to use explicit URLs (write in-place instead of putting into a variable).

fixed

profiles/tests/test_models.py
Krymets: At the moment, I consider it unacceptable to implement logging + tokens in each test.
If the tokens are removed later, I will rewrite the tests

PR #148, opened 2023-08-24
test for crud-operations profile

BelousSofiya:


profiles/tests/test_crud_profile.py
```diff
@@ -0,0 +1,302 @@
...
+            comp_is_startup=True,
+            comp_registered=False
+        )
+        cls.test_profile2 = Profile.objects.create(
+            person=cls.test_person2,
+            comp_registered=True,
+            comp_is_startup=False
+        )
+
+        Category.objects.create(category_id=1,
```
Shushunya, line 49: "Producer" is a type of Activitty, not Category, as I understand. Better change that to `cheese` or something.

BelousSofiya: Thank you. Done

profiles/views.py
```diff
@@ -84,12 +86,27 @@ class ProfileDetail(RetrieveUpdateDestroyAPIView):
...
     permission_classes = (IsAuthenticated,)
 
-    def get_queryset(self):
+    def get_queryset(self, pk=None):
         user_id = self.request.user.id
         if self.request.method == "DELETE":
             return Profile.objects.filter(person_id=user_id, is_deleted=False)
         if self.request.method == 'GET':
             return Profile.objects.filter(is_deleted=False)
+        if self.request.method == 'PUT':
```
Shushunya, line 95: You can combine these two, since they return the same objects. I suggest `if self.request.method in ["PUT", "PATCH"]`

BelousSofiya: You are right. Thank you. Done.

mehalyna (approved):

Issue #142, opened 2023-08-22
Implement the functionality of view phone number and email address


VovaD199:
#61 

https://www.figma.com/file/gMu88CJcuaJSSoTGC5ZYfD/Forum?type=design&node-id=966-61572&mode=design&t=CLnKZAV2tl0JXdDw-4

PR #150, opened 2023-08-25
#142 add viewed company

Shushunya:


Lvyshnevska (approved):

profiles/models.py
```diff
@@ -52,3 +52,11 @@ class SavedCompany(models.Model):
     company = models.ForeignKey(Profile, on_delete=models.CASCADE, related_name='saved_list')
     user = models.ForeignKey(CustomUser, on_delete=models.CASCADE, related_name='saved_list_items')
     added_at = models.DateTimeField(auto_now_add=True)
+
+
+class ViewedCompany(models.Model):
+    class Meta:
```
ProDG, line 58: Put it after all fields, please.

profiles/serializers.py
```diff
@@ -41,3 +41,11 @@ class Meta:
         model = SavedCompany
         fields = ('id', 'user', 'company', 'comp_official_name', 'comp_region', 'comp_common_info', 'comp_phone_number',
                   'comp_EDRPOU', 'comp_year_of_foundation', 'comp_address', 'startup_idea', 'added_at')
+
+
+class ViewedCompanySerializer(serializers.ModelSerializer):
+    # profile_id = ProfileSerializer()
+    # user_id =
```
ProDG, line 48: Remove commented lines.

profiles/serializers.py
```diff
@@ -41,3 +41,11 @@ class Meta:
...
+    class Meta:
+        model = ViewedCompany
+        fields = "__all__"
```
ProDG, line 51: I strongly recommend to not use `__all__`. When you go this way, you can later add some field to the model which you don't actually want to be exposed by API, and it will be exposed.

profiles/tests/test_filter_profiles.py
```diff
@@ -13,7 +13,7 @@ def setUpTestData(cls):
         test_categories = []
 
         # setup activities
-        HEADER_ACTIVITIES = ["producer", "importer", "retail", "HORACE"]
+        HEADER_ACTIVITIES = ["producer", "importer", "retail", "HORECA"]
```
ProDG, line 16: I'd recommend to:
1. Use constants instead of 'magic strings'.
2. Be consistent in letters case. If all of the strings in this list are codes, make them either all lowercase or all uppercase, but not mixed.

profiles/tests/test_filter_profiles.py
```diff
@@ -129,14 +129,14 @@ def test_get_all_profiles_authorized_filter_activity_retail(self):
                        3 in response.data[i]["comp_activity"]]
         self.assertEqual(len(retail_type), filtered_len)
 
-    def test_get_all_profiles_authorized_filter_activity_HORACE(self):
-        response = self.client.get(f"/api/profiles/?activity_type=HORACE")
+    def test_get_all_profiles_authorized_filter_activity_HORECA(self):
```
ProDG, line 132: In method names it _definitely_ must be lowercase.

profiles/tests/test_filter_profiles.py
```diff
@@ -129,14 +129,14 @@ def test_get_all_profiles_authorized_filter_activity_retail(self):
...
+        response = self.client.get(f"/api/profiles/?activity_type=HORECA")
         filtered_len = len(response.data)
         self.assertEqual(200, response.status_code)
         response = self.client.get("/api/profiles/")
-        horace_type = [response.data[i]["comp_activity"] for i in range(len(response.data)) if
+        HORECA_type = [response.data[i]["comp_activity"] for i in range(len(response.data)) if
                        4 in response.data[i]["comp_activity"]]
-        self.assertEqual(len(horace_type), filtered_len)
+        self.assertEqual(len(HORECA_type), filtered_len)
```
ProDG, line 139: I _strongly_ recommend to do _explicit_ checks.

I guess you know what exactly number you expect to have here. Then use this number in the assert.

Why? Because this test will pass in case when it is actuall broken (let's say, you expected here to have number 2 both in expected and actual results, but got such errors, which leads to 0 in both cases).

It's really, really important — dont write asserts list `self.assertEqual(code_calculating_one_value(), code_calculating_another_value())`. Write asserts like `self.assertEqual(42, code_calculating_value())`, where 42 is the exact value you expect to have.

It's better to not have a unittest at all, than to have one, which cofuses you with wrong feeling like 'I have coverage here'.

profiles/tests/test_viewed_company.py
```diff
@@ -0,0 +1,77 @@
...
+        self.test_person = CustomUser.objects.create_user(
+            person_email="test1@test.com",
+            password="Testing01",
+            person_name="test1",
+            person_surname="test",
+            is_active=True
+        )
+        test_person2 = CustomUser.objects.create_user(
+            person_email="test2@test.com",
+            password="Testing01",
```
ProDG, line 18: I doubt if passwords are set like this.

ProDG: ... and emphasize once again — use factories.

profiles/tests/test_viewed_company.py
```diff
@@ -0,0 +1,77 @@
...
+            comp_registered=True,
+            comp_is_startup=False
+        )
+        self.test_profile3 = Profile.objects.create(
+            person=test_person3,
+            comp_registered=True,
+            comp_is_startup=False
+        )
+
+        self.client = APIClient()
```
ProDG, line 46: Please DON'T do like this.

If the class is derived from `APITestCase`, you already have `self.client` properly set.

profiles/tests/test_viewed_company.py
```diff
@@ -0,0 +1,77 @@
...
+        self.client.force_authenticate(self.test_person)
+        response = self.client.post("/api/viewed-list/", data={"profile_id": self.test_profile.profile_id})
+        self.assertEqual(400, response.status_code)
+
+    def test_get_viewed_companies_unauthorized(self):
+        response = self.client.post("/api/viewed-list/")
+        self.assertEqual(401, response.status_code)
+
+    def test_get_viewed_companies_authorized(self):
+        from datetime import datetime
```
ProDG, line 69: Put at the beginning of the file.

profiles/views.py
```diff
@@ -102,3 +102,26 @@ def destroy(self, request, pk=None):
...
+    def create(self, request, *args, **kwargs):
+        user_id = self.request.user.id
+        company_id = request.data.get("profile_id")
+        company = get_object_or_404(Profile, profile_id=company_id)
+        if company.person.pk != user_id:
+            serializer = self.serializer_class(data={"user": user_id, "company": company_id})
+            if serializer.is_valid():
+                serializer.save()
+                return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(status=status.HTTP_400_BAD_REQUEST)
```
ProDG, line 127: Simple but important rule (please share with the rest of the team).

If you're 'handcrafting' response in DRF code — 99% you're doing something wrong.

You must not do `return Response(...)`. 

In the line 127 you return `400 BAD REQUEST` but there is no data, what exactly is wrong with the request (because by handcrafting request you just break everything which is already implemented in DRF, in a proper way).

profiles/views.py
```diff
@@ -102,3 +102,26 @@ def destroy(self, request, pk=None):
...
+
+    def list(self, request, *args, **kwargs):
+        user_id = self.request.user.id
+        viewed_companies = ViewedCompany.objects.filter(user=user_id)
+        serializer = self.serializer_class(viewed_companies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    def create(self, request, *args, **kwargs):
+        user_id = self.request.user.id
+        company_id = request.data.get("profile_id")
```
ProDG, line 120: Don't do manual manipulations with payload, unless there is no other way to achieve something.

This is work of serializer.

Shushunya: Moved validation to serializer, removed overriding of `create` in the view.

profiles/views.py
```diff
@@ -102,3 +102,26 @@ def destroy(self, request, pk=None):
...
+class ViewedCompanyList(ListCreateAPIView):
+    queryset = ViewedCompany.objects.all()
+    serializer_class = ViewedCompanySerializer
+    permission_classes = (IsAuthenticated, )
+
+    def list(self, request, *args, **kwargs):
+        user_id = self.request.user.id
+        viewed_companies = ViewedCompany.objects.filter(user=user_id)
+        serializer = self.serializer_class(viewed_companies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
Shushunya: Done.

profiles/models.py
Shushunya: Done.

profiles/serializers.py
Shushunya: Done.

profiles/serializers.py
Shushunya: Done.

profiles/tests/test_filter_profiles.py
Shushunya: Done.

profiles/tests/test_viewed_company.py
Shushunya: Removed.

profiles/tests/test_viewed_company.py
Shushunya: Removed it at all. It was unnecessary, missed it before push.

profiles/views.py
Shushunya: Now I do not override `create`.

mehalyna (approved):

Issue #136, opened 2023-08-22
Implement the functionality of view 'Про нас'.


VovaD199:
#75 need implement

PR #168, opened 2023-08-31
#136 about us navigation

Shushunya:
User story:

As a user,
I want to review the information about platform
So that I understand what this platform offers and if I am interested in it.

Given I am on the platform
When I click on 'Про нас' button
Then I am redirected to 'Про нас' section on the main page

### DIFF
--- a/FrontEnd/src/components/HeaderFooter/footer/top/FooterNavigation.jsx
+++ b/FrontEnd/src/components/HeaderFooter/footer/top/FooterNavigation.jsx
@@ -1,80 +1,100 @@
 import css from "./FooterNavigation.module.css";
+import { HashLink } from "react-router-hash-link";
 
 const PAGE_NAVIGATION_LINKS = [
-    {
-        id: "pnl0",
-        title: "Головна",
-        link: "#"
-    },
-    {
-        id: "pnl1",
-        title: "Компанії",
-        link: "#"
-    },
-    {
-        id: "pnl2",
-        title: "Стартапи",
-        link: "#"
-    },
-    {
-        id: "pnl3",
-        title: "Про нас",
-        link: "#"
-    }
-]
+  {
+    id: "pnl0",
+    title: "Головна",
+    link: "/",
+  },
+  {
+    id: "pnl1",
+    title: "Компанії",
+    link: "#",
+  },
+  {
+    id: "pnl2",
+    title: "Стартапи",
+    link: "#",
+  },
+  {
+    id: "pnl3",
+    title: "Про нас",
+    link: "/#about-us",
+  },
+];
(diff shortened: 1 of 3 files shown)


FrontEnd/src/components/HeaderFooter/footer/top/FooterNavigation.jsx
```diff
@@ -1,80 +1,100 @@
...
+  },
+  {
+    id: "pnl2",
+    title: "Стартапи",
+    link: "#",
+  },
+  {
+    id: "pnl3",
+    title: "Про нас",
+    link: "/#about-us",
```
popovycholeg, line 23: no need  to add `/` before anchor

FrontEnd/src/components/HeaderFooter/footer/top/FooterNavigation.jsx
```diff
@@ -1,80 +1,100 @@
...
+            className={css["navigation-content-section__text"]}
+            to={element.link}
+          >
+            {element.title}
+          </HashLink>
+        ))}
+      </div>
+      <div className={css["navigation-content-section"]}>
+        {SERVICES_LINKS.map((element) => (
+          <a
```
popovycholeg, line 77: Since we use react router for the entire app, its better to use `Link` from the `react-router` to utilize full power of SPA

FrontEnd/src/components/HeaderFooter/header/navbar/Menu.jsx
```diff
@@ -10,17 +10,17 @@ const MENU_LINKS = [
...
         id: "m2",
         title: "Стартапи",
-        link: "#"
+        link: "/"
     },
     {
         id: "m3",
         title: "Про нас",
-        link: "#"
+        link: "/#about-us"
```
popovycholeg, line 23: same no need to put `/` with `#hash-fragment`

popovycholeg (approved):
LGTM! Added few minor comments

FrontEnd/src/components/HeaderFooter/footer/top/FooterNavigation.jsx
Shushunya: As far as I know, `Link` doesn't support anchor links. That's why I decided to use `HashLink`.

RedSSD (approved):

FrontEnd/src/components/HeaderFooter/footer/top/FooterNavigation.jsx
Shushunya: This is the anchor for the main page section and without slash it tries to find the anchor on the other page. So this slash is necessary since we need to have this functionality working from any place.

Issue #159, opened 2023-08-30
Investigate Regions In Django


ddianaoo:
django-cities-light

PR #180, opened 2023-09-06
#159 regions

ddianaoo:
I investigated django-cities-light library, but decided not to use it in our project. So I created a table of regions, with a ManyToMany relation to a Profile.


profiles/migrations/0005_alter_profile_comp_region.py
```diff
@@ -13,6 +13,6 @@ class Migration(migrations.Migration):
...
...=models.CharField(choices=[('Kyiv', 'Київ'), ('Vinnytsia region', 'Вінницька область'), ('Volyn region', 'Волинська область'), ('Dnipro region', 'Дніпропетровська область'), ('Donetsk region', 'Донецька область'), ('Zhytomyr region', 'Житомирська область'), ('Zakarpattia region', 'Закарпатська область'), ('Zaporizhzhia region', 'Запорізька область'), ('IvanoFrankivsk region', 'Івано-Франківська область'), ('Kyiv region', 'Київська область'), ('Kirovohrad region', 'Кіровоградська область'), ('Crimea', 'Автономна Республіка Крим'), ('Luhansk region', 'Луганська область'), ('Lviv region', 'Львівська область'), ('Mykolaiv region', 'Миколаївська область'), ('Odesa region', 'Одеська область'), ('Poltava region', 'Полтавська область'), ('Rivne region', 'Рівненська область'), ('Sevastopol', 'Севастополь'), ('Sumy region', 'Сумська область'), ('Ternopil region', 'Тернопільська область'), ('Kharkiv region', 'Харківська область'), ('Kherson region', 'Херсонська область'), ('Khmelnytskyi region', 'Хмельницька область'), ('Cherkasy region', 'Черкаська область'), ('Chernivtsi region', 'Чернівецька область'), ('Chernihiv region', 'Чернігівська область')], default=None, max_length=128, null=True),
```
Shushunya, line 16: I think it's better to use `choices=Region.choices` instead of duplicating all the values from it.

GeorgiyII (approved):

GeorgiyII:
`VinnytsiaReg = 'Vinnytsia region', "Вінницька область"`

Such code is not python style. In the python we use snake-case for variables or methods, camel-case we use only for classes.

So it will be better to write such rows like:
`vinnytsia_region = 'Vinnytsia region', "Вінницька область"` or
`VINNYTSA_REGION = 'Vinnytsia region', "Вінницька область"`

Shushunya (approved):

Issue #171, opened 2023-09-03
Implement the pagination (back-end) for the list of Companies page according to the mockups on Figma


VovaD199:
Implement the pagination for the list of Companies page according to the mockups on Figma (back-end)

https://www.figma.com/file/gMu88CJcuaJSSoTGC5ZYfD/Forum?type=design&node-id=1788-13571&mode=design&t=s21pv44c0lNTs1BS-4

PR #185, opened 2023-09-07
#171 implement pagination

Lvyshnevska:
Pagination for Companies list was implemented on the back-end. Profile tests were updated according to new response data (taking into account pagination functionality). Besides, imports order were fixed in some files.


forum/settings.py
```diff
@@ -11,6 +11,7 @@
 """
 
 from datetime import timedelta
+from os import nice
```
ProDG, line 14: Is it used?

profiles/migrations/0005_alter_profile_options_alter_savedcompany_options_and_more.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='profile',
+            options={'ordering': ['profile_id']},
+        ),
+        migrations.AlterModelOptions(
+            name='savedcompany',
+            options={'ordering': ['company_id']},
```
ProDG, line 19: I really, really discourage you of using `Meta.ordering` on models.

While it works nicely in cases when project is small and there is not a lot of records in DB, it brings just HUGE amount of issues on bigger projects.

Ordering must be applied explicitly, _when_ needed and _where_ needed.

ProDG: To give you more context. This ordering will be applied on each DB request to the model, doesn't matter if there is a sense to do ordering (which, by the way, is not 'for free', it's pretty heavy operation on DB level).

I'm not 100% sure, but it's possible that with `Meta.ordering` on model you'll have issues trying to implement ordering in DRF filters etc.

So, just don't use this.

forum/settings.py
Lvyshnevska: o_O It's crazy VS Code. I'll remove it.

ProDG (approved):

GeorgiyII (approved):

Issue #172, opened 2023-09-03
Implement the filter "Усі\Збережені" for the list of Companies Page


VovaD199:
Implement the filter "Усі\Збережені" for the list of Companies Page

https://www.figma.com/file/gMu88CJcuaJSSoTGC5ZYfD/Forum?type=design&node-id=1788-13571&mode=design&t=s21pv44c0lNTs1BS-4

PR #191, opened 2023-09-10
#172 All/saved companies functionality

Lvyshnevska:
All/saved companies functionality implemented. Field "saved" added to ProfileSerializer. It returns True, if company is in saved companies list of user. User's Companies saved list  returns by passing query param by "api/profiles/" endpoint. Due to this SavedCompanies functionality and test updated a little.

### DIFF
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -1,7 +1,6 @@
 from rest_framework import serializers
 
 from authentication.models import CustomUser
-
 from .models import Profile, Activity, Category, SavedCompany, ViewedCompany
 
 
@@ -21,13 +20,13 @@ class ProfileSerializer(serializers.ModelSerializer):
     activity = ActivitySerializer(many=True, read_only=True)
     category = CategorySerializer(many=True, read_only=True)
     person = serializers.PrimaryKeyRelatedField(queryset=CustomUser.objects.all())
-    saved = serializers.SerializerMethodField()
+    is_saved = serializers.SerializerMethodField()
 
     class Meta:
         model = Profile
         fields = "__all__"
 
-    def get_saved(self, obj):
+    def get_is_saved(self, obj):
         user = self.context["request"].user
         if user.is_authenticated:
             saved_companies = obj.saved_list.filter(user=user)
@@ -86,3 +85,6 @@ class ViewedCompanySerializer(serializers.ModelSerializer):
         if company.person == user:
             raise serializers.ValidationError({"error": "You can not view your company."})
         return attrs
+
+class QueryParamSerializer(serializers.Serializer):
+    filters = serializers.CharField(required=True)
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -1,3 +1,4 @@
+import re
 from rest_framework import serializers
 
 from authentication.models import CustomUser
@@ -86,5 +87,9 @@ class ViewedCompanySerializer(serializers.ModelSerializer):
             raise serializers.ValidationError({"error": "You can not view your company."})
         return attrs
 
-class QueryParamSerializer(serializers.Serializer):
+class FiltersQueryParamSerializer(serializers.Serializer):
     filters = serializers.CharField(required=True)
+
+    def to_representation(self, instance):
+        representation = {"instance": "instance"}
+        return representation
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -21,11 +21,19 @@ class ProfileSerializer(serializers.ModelSerializer):
     activity = ActivitySerializer(many=True, read_only=True)
     category = CategorySerializer(many=True, read_only=True)
     person = serializers.PrimaryKeyRelatedField(queryset=CustomUser.objects.all())
+    saved = serializers.SerializerMethodField()
 
     class Meta:
         model = Profile
         fields = "__all__"
 
+    def get_saved(self, obj):
+        user = self.context["request"].user
+        if user.is_authenticated:
+            saved_companies = obj.saved_list.filter(user=user)
+            return saved_companies.exists() 
+        return False
+    
 
 class ProfileDetailSerializer(serializers.ModelSerializer):
     activity = ActivitySerializer(many=True, read_only=True)
--- a/profiles/views.py
+++ b/profiles/views.py
@@ -7,8 +7,11 @@ from rest_framework import status
 from forum.pagination import ForumPagination
 from .models import SavedCompany, Profile, ViewedCompany
 from .serializers import (SavedCompanySerializer, ProfileSerializer, ViewedCompanySerializer,
-                          ProfileSensitiveDataROSerializer, ProfileDetailSerializer)
+                          ProfileSensitiveDataROSerializer, ProfileDetailSerializer, QueryParamSerializer)
 from .permissions import UserIsProfileOwnerOrReadOnly, SavedCompaniesListPermission
+
+
+FILTERS= {"saved": "is_saved"}
 
 
 class SavedCompaniesCreate(CreateAPIView):
@@ -60,7 +63,7 @@ class ProfileList(ListCreateAPIView):
     pagination_class = ForumPagination
 
     def get_queryset(self):
-        companies = self.request.query_params.get("companies")
+        query_params_serializer = QueryParamSerializer(data=self.request.query_params)
         company_type = self.request.query_params.get("company_type")
         activity_type = self.request.query_params.get("activity_type")
         HEADER_ACTIVITIES = ["producer", "importer", "retail", "horeca"]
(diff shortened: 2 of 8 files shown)


ProDG (approved):

mehalyna (approved):

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
...
+        queryset = Profile.objects.filter(is_deleted=False).order_by("profile_id")
+       
+        if userid:
             try:
-                userid = self.request.query_params.get('userid')
-                return Profile.objects.filter(person_id=userid)
+                return queryset.filter(person_id=userid)
             except ValueError:
-                raise ValidationError(detail='Bad request')
+                pass
```
GeorgiyII, line 78: query set in this situation can't raise ValueError, this try-except is not needed here

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
...
     serializer_class = ProfileSerializer
-    permission_classes = (IsAuthenticatedOrReadOnly, )
+    permission_classes = [IsAuthenticatedOrReadOnly & SavedCompaniesListPermission]
     pagination_class = ForumPagination
 
     def get_queryset(self):
+        query_params_serializer = QueryParamSerializer(data=self.request.query_params)
         company_type = self.request.query_params.get("company_type")
         activity_type = self.request.query_params.get("activity_type")
+        userid = self.request.query_params.get("userid")
```
GeorgiyII, line 69: please create `user_id` variable instead of `userid`

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
      include_all: bool.
     """
     serializer_class = ProfileSerializer
-    permission_classes = (IsAuthenticatedOrReadOnly, )
+    permission_classes = [IsAuthenticatedOrReadOnly & SavedCompaniesListPermission]
     pagination_class = ForumPagination
 
     def get_queryset(self):
+        query_params_serializer = QueryParamSerializer(data=self.request.query_params)
```
GeorgiyII, line 66: I told you about FilterQueryParamsSerializer, so you should have something like this here:

filters = FilterQueryParamsSerializer(data=self.request.query_params.get("filters"))

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
...
         if activity_type in HEADER_ACTIVITIES:
-            return Profile.objects.filter(comp_activity__name=activity_type).order_by("profile_id")
-
-        return Profile.objects.filter(is_deleted=False).order_by("profile_id")
+            return queryset.filter(comp_activity__name=activity_type)
+        if query_params_serializer.is_valid():
+            data = query_params_serializer.validated_data
+            filters = data.get("filters")
+            if filters == FILTERS["saved"]:
+                return queryset.filter(saved_list__user=self.request.user)
```
GeorgiyII, line 89:
```suggestion
        if filters.is_valid():
            if filters.get("saved"):
                queryset = queryset.filter(saved_list__user=self.request.user)
```

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
...
-            return Profile.objects.filter(comp_is_startup=True).order_by("profile_id")
+            return queryset.filter(comp_is_startup=True)
         elif company_type == "company":
-            return Profile.objects.filter(comp_registered=True).order_by("profile_id")
+            return queryset.filter(comp_registered=True)
         if activity_type in HEADER_ACTIVITIES:
-            return Profile.objects.filter(comp_activity__name=activity_type).order_by("profile_id")
-
-        return Profile.objects.filter(is_deleted=False).order_by("profile_id")
+            return queryset.filter(comp_activity__name=activity_type)
```
GeorgiyII, line 84: Too many returns in one function, especially if you will want to use multiple filters

queryset = queryset.filter(comp_activity__name=activity_type)

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
...
+                return queryset.filter(person_id=userid)
             except ValueError:
-                raise ValidationError(detail='Bad request')
+                pass
         if company_type == "startup":
-            return Profile.objects.filter(comp_is_startup=True).order_by("profile_id")
+            return queryset.filter(comp_is_startup=True)
         elif company_type == "company":
-            return Profile.objects.filter(comp_registered=True).order_by("profile_id")
+            return queryset.filter(comp_registered=True)
```
GeorgiyII, line 82: queryset = queryset.filter(comp_registered=True)

profiles/views.py
```diff
@@ -61,28 +59,35 @@ class ProfileList(ListCreateAPIView):
...
             try:
-                userid = self.request.query_params.get('userid')
-                return Profile.objects.filter(person_id=userid)
+                return queryset.filter(person_id=userid)
             except ValueError:
-                raise ValidationError(detail='Bad request')
+                pass
         if company_type == "startup":
-            return Profile.objects.filter(comp_is_startup=True).order_by("profile_id")
+            return queryset.filter(comp_is_startup=True)
```
GeorgiyII, line 80: queryset = queryset.filter(comp_is_startup=True)

profiles/views.py
Lvyshnevska: But if we pass as "userid" value that not an integer, ValueError occurs. So try/except lets us to avoid this error and to return queryset. Or there is another way to solve this issue?

profiles/views.py
Lvyshnevska: Fixed

profiles/views.py
Lvyshnevska: Fixed

profiles/views.py
Lvyshnevska: I'll improve it next sprint

profiles/serializers.py
```diff
@@ -78,3 +85,6 @@ def validate(self, attrs):
         if company.person == user:
             raise serializers.ValidationError({"error": "You can not view your company."})
         return attrs
+
+class QueryParamSerializer(serializers.Serializer):
+    filters = serializers.CharField(required=True)
```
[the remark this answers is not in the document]
Lvyshnevska: I'll improve it next sprint

profiles/serializers.py
```diff
@@ -1,7 +1,7 @@
+import re
```
ProDG, line 1: Is it used? If not — remove the import.

profiles/serializers.py
```diff
@@ -21,11 +21,19 @@ class ProfileSerializer(serializers.ModelSerializer):
...
     class Meta:
         model = Profile
         fields = "__all__"
 
+    def get_is_saved(self, obj):
+        user = self.context["request"].user
+        if user.is_authenticated:
+            saved_companies = obj.saved_list.filter(user=user)
+            return saved_companies.exists() 
+        return False
```
ProDG, line 35: In case someone will decide to use this serialize in List endpoint, we'll have a trouble. 

For _each_ of the profile, this code (containing call to DB) will be executed, so there will be N+1 problem.

If we're sure that the serializer will be used only for details endpoint (and in this case adding `Details` word into the name would be helpful), we can keep it as is.

If not (let's say, we know for sure that it will be used for List endpoint), we can do the next trick:

1. Retrieve PKs of all comanies, saved by current user.
2. Put this data as a list (or, better — frozenset) into serializer context (by overriding `get_serializer_context()`)
3. This method will contain a single string of code, something like `return obj.pk in self.context['saved_companies_pks']`
4. .. and also, you can combine these approaches — check if there is `saved_companies_pks` in context, and do the next steps based on the result.

profiles/serializers.py
```diff
@@ -78,3 +86,10 @@ def validate(self, attrs):
         if company.person == user:
             raise serializers.ValidationError({"error": "You can not view your company."})
         return attrs
+
+class FiltersQueryParamSerializer(serializers.Serializer):
+    filters = serializers.CharField(required=True)
+
+    def to_representation(self, instance):
+        representation = {"instance": "instance"}
+        return representation
```
ProDG, line 95: I have strong feeling that something is very wrong here. 

Serializers have nothing to do with query parameters.

profiles/views.py
```diff
@@ -62,28 +56,35 @@ class ProfileList(ListCreateAPIView):
      include_all: bool.
     """
     serializer_class = ProfileSerializer
-    permission_classes = (IsAuthenticatedOrReadOnly, )
+    permission_classes = [IsAuthenticatedOrReadOnly & SavedCompaniesListPermission]
     pagination_class = ForumPagination
 
     def get_queryset(self):
+        filters = FiltersQueryParamSerializer(data=self.request.query_params)
```
ProDG, line 63: Nope. It _must_ be implemented using this feature of DRF:

https://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend

ProDG: You can merge this PR if a task for fixing this is immediately created. Or fix it before merging.

PR #194, opened 2023-09-11
93 admin panel users

RedSSD:


administration/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+from django.contrib.auth import get_user_model
+from django.core.exceptions import ObjectDoesNotExist
+from rest_framework.serializers import SerializerMethodField
+
+User = get_user_model()
+
+
+class AdminUserSerializer(UserSerializer):
+
+    phone_number = SerializerMethodField('get_phone_number')
```
ProDG, line 11: `'get_phone_number'` string is redundand and should be removed (as it matches the name expected for the method by default).

administration/views.py
```diff
@@ -1,3 +1,29 @@
...
+
+class UserDetailView(RetrieveUpdateDestroyAPIView):
+    """
+    Retrieve, update or delete a user.
+    """
+    permission_classes = [IsAuthenticated, IsAdminUser]
+    serializer_class = AdminUserSerializer
+
+    def get_queryset(self):
+        return CustomUser.objects.filter(is_superuser=False).order_by('id')
```
ProDG, line 29: Should be just `queryset = ...`.

`get_queryset()` should be used when we need some logic which requires access to the view's `self`.

ProDG (approved):

mehalyna (approved):

PR #195, opened 2023-09-12
#92 admin panel companies

whooaami:


administration/urls.py
```diff
@@ -1,9 +1,11 @@
-from django.urls import path
-from .views import UsersListView, UserDetailView
+from django.urls import path, re_path, include
+from .views import AdminProfileList, AdminProfileDetail
```
GeorgiyII, line 2: please add one pass line between libraries and custom modules

and please use one style of imports in whole project, here you have relative import, in previous file you have full path import

administration/views.py
```diff
@@ -1,25 +1,26 @@
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
-from rest_framework.generics import ListAPIView, RetrieveUpdateDestroyAPIView
+from rest_framework.generics import ListCreateAPIView, RetrieveUpdateDestroyAPIView
+from rest_framework.response import Response
+from profiles.models import Profile
```
GeorgiyII, line 4: please add one pass line between libraries and custom modules

administration/serializers.py
```diff
@@ -1,31 +1,10 @@
...
-            "person_surname",
-            "person_email",
-            "phone_number",
-            "last_login",
-            "is_active",
-            "is_staff"
-        )
-        read_only_fields = ("phone_number", "person_email")
+class AdminCompanySerializer(serializers.ModelSerializer):
+    person = serializers.PrimaryKeyRelatedField(queryset=CustomUser.objects.all())
```
GeorgiyII, line 7: what field you wont to have here in person attribute?

if you want to have whole object maybe it would be better to use one more model serializer to get CustomUser?

Like in documentation: https://www.django-rest-framework.org/api-guide/relations/#nested-relationships

administration/serializers.py
```diff
@@ -1,31 +1,10 @@
-from djoser.serializers import UserSerializer
-from django.contrib.auth import get_user_model
-from django.core.exceptions import ObjectDoesNotExist
-from rest_framework.serializers import SerializerMethodField
+from rest_framework import serializers
+from authentication.models import CustomUser
```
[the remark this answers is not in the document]
whooaami: Fixed

administration/urls.py
whooaami: Fixed

administration/views.py
whooaami: Fixed

administration/serializers.py
```diff
@@ -1,31 +1,11 @@
...
-            "last_login",
-            "is_active",
-            "is_staff"
-        )
-        read_only_fields = ("phone_number", "person_email")
+class AdminCompanySerializer(serializers.ModelSerializer):
+    person = serializers.PrimaryKeyRelatedField(queryset=CustomUser.objects.all())
+    class Meta:
+        model = Profile
+        fields = "__all__"
```
ProDG, line 11: Again, I don't recommend using `__all__`, explicit fields enumeration is safer.

administration/urls.py
```diff
@@ -1,9 +1,12 @@
...
 
 app_name = "administration"
 
 urlpatterns = [
-    path('users/', UsersListView.as_view(), name='users-list'),
-    path('users/<pk>/', UserDetailView.as_view(), name='user'),
+    path("auth/", include('djoser.urls')),
+    re_path(r"^auth/", include('djoser.urls.authtoken')),
+    path('profiles/', AdminProfileList.as_view(), name='profile-list'),
+    path('profiles/<pk>', AdminProfileDetail.as_view(), name='profile-detail'),
```
ProDG, line 11: Use trailing `/` in all URLs!

administration/urls.py
```diff
@@ -1,9 +1,12 @@
...
+
+from administration.views import AdminProfileList, AdminProfileDetail
 
 app_name = "administration"
 
 urlpatterns = [
-    path('users/', UsersListView.as_view(), name='users-list'),
-    path('users/<pk>/', UserDetailView.as_view(), name='user'),
+    path("auth/", include('djoser.urls')),
+    re_path(r"^auth/", include('djoser.urls.authtoken')),
```
ProDG, line 9: Is this correct? Do we need `djoser` for authentication, we have strong reasons to consider it better/simpler than Django's built-in users authentication with Django sessions and cookies?

administration/views.py
```diff
@@ -1,25 +1,27 @@
...
+    permission_classes = (IsAuthenticated, IsAdminUser)
+    serializer_class = AdminCompanySerializer
+    queryset = Profile.objects.filter(is_deleted=False)
+    
+    def update(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
```
ProDG, line 27: Do we need this method? It looks like it does almost exactly the same what does built-in DRFs implementation. Probably only making all updates 'partial' (and I'd believe it's not a good idea to do this).

Do we have to do `return Response()` here?

profiles/models.py
```diff
@@ -69,6 +69,7 @@ class Profile(models.Model):
     banner_image = models.ImageField(validators=[validate_image_format, validate_image_size], null=True)
 
     is_deleted = models.BooleanField(default=False)
+    is_approved = models.BooleanField(default=False)
```
ProDG, line 72: What is a meaning of this field? I'm not aware about all updates of requirements, do we have profile approval included?

whooaami: At the meeting, we discussed this field. That is, the administrator checks the EDRPOU and if it is valid, then sets is_approved=True.
We don't have exact requirements for the admin panel.
If this field is not okay, I can remove it from the code.

administration/urls.py
whooaami: We also discussed this and decided to keep the authentication that has already been implemented.
Again, we need the requirements for the admin panel so that we can do everything correctly.

administration/urls.py
whooaami: Fixed

administration/serializers.py
whooaami: Fixed

administration/serializers.py
whooaami: Fixed

GeorgiyII (approved):

administration/views.py
whooaami: Remove this method.

RedSSD (approved):

VovaD199 (approved):

forum/settings.py
```diff
@@ -178,6 +178,8 @@
     'PASSWORD_RESET_CONFIRM_URL': '#/password/reset/confirm/{uid}/{token}',
     'PASSWORD_RESET_CONFIRM_RETYPE': True,
     'PASSWORD_RESET_SHOW_EMAIL_NOT_FOUND': True,
+    'ACTIVATION_URL': '#/activate/{uid}/{token}',
```
ProDG, line 181: Not sure, but probably here we also should use trailing slashes in URLs.

ProDG (approved):

PR #207, opened 2023-09-14
#189managing categoryand activity

BelousSofiya:
CRUD for Category and Activity + tests.
Every user (even not authenticated) has opportunity to get list of activities and categories. But only admins can create, put, patch, delete Category and Activity.


profiles/permissions.py
```diff
@@ -15,3 +15,10 @@ def has_permission(self, request, view):
         if request.query_params.get("filters") == "is_saved":
             return request.user.is_authenticated
         return True
+
+
+class UserIsAdmin(BasePermission):
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
+        return request.user.is_staff
```
Shushunya, line 24: There's a DRF built-in permission [`IsAdminUser`](https://www.django-rest-framework.org/api-guide/permissions/#isadminuser). I guess it would be better if you've had something like this:
```suggestion
class IsAdminOrReadOnly(IsAdminUser):
    def has_permission(self, request, view):
        is_admin = super().has_permission(request, view)
        return request.method in SAFE_METHODS or is_admin
```

profiles/tests/test_activity.py
```diff
@@ -1,18 +1,126 @@
...
+            person_email="testactivity@testuser.com",
+            password="Testing01",
+            person_name="testactivity",
+            person_surname="user",
+            is_active=True
+        )
 
     def tearDown(self) -> None:
         objects_to_delete = Activity.objects.all()
         objects_to_delete.delete()
```
Shushunya, line 31: There's no need to clean up DB in the `tearDown` method since it's done automatically after each test method.

profiles/tests/test_category.py
```diff
@@ -8,11 +10,116 @@ def setUp(self) -> None:
...
+            person_email="testcategory@testuser.com",
+            password="Testing01",
+            person_name="testcategory",
+            person_surname="user",
+            is_active=True
+        )
 
     def tearDown(self) -> None:
         objects_to_delete = Category.objects.all()
         objects_to_delete.delete()
```
Shushunya, line 32: Same as in `test_activity.py`:

> There's no need to clean up DB in the `tearDown` method since it's done automatically after each test method.

profiles/views.py
```diff
@@ -135,18 +135,34 @@ def get_queryset(self):
...
     serializer_class = RegionSerializer
     queryset = Region.choices
+
+
+class CategoryDetail(RetrieveUpdateDestroyAPIView):
+    serializer_class = CategorySerializer
+    permission_classes = (IsAdminUser,)
+
+    def get_queryset(self, pk=None):
+        return Category.objects.all()
```
Shushunya, line 160: There's no need to override `get_queryset` method here. Just add `queryset`.
```suggestion
    queryset = Category.objects.all()
```

profiles/views.py
```diff
@@ -135,18 +135,34 @@ def get_queryset(self):
...
+
+
+class ActivityDetail(RetrieveUpdateDestroyAPIView):
+    serializer_class = ActivitySerializer
+    permission_classes = (IsAdminUser,)
+
+    def get_queryset(self, pk=None):
+        return Activity.objects.all()
```
Shushunya, line 168: Same as in `CategoryDetail` above.

> There's no need to override `get_queryset` method here. Just add `queryset`.

profiles/urls.py
```diff
@@ -12,6 +13,8 @@
     path('saved-list/<pk>/', SavedCompaniesDestroy.as_view(), name='saved_companies_destroy'),
     path('viewed-list/', ViewedCompanyList.as_view(), name='viewed_company_list'),
     path('regions/', RegionListView.as_view(), name='region-list'),
-    path('category-list/', CategoryList.as_view(), name='category_list'),
-    path('activity-list/', ActivityList.as_view(), name='activity_list'),
+    path('category/', CategoryList.as_view(), name='category_list'),
+    path('activity/', ActivityList.as_view(), name='activity_list'),
+    path('category/<pk>', CategoryDetail.as_view(), name='category-detail'),
+    path('activity/<pk>', ActivityDetail.as_view(), name='activity-detail'),
```
Shushunya, line 19: Previously we used plural forms of the collection name. And it's a common practice for REST APIs in general. Here's [an article](https://www.freecodecamp.org/news/rest-api-best-practices-rest-endpoint-design-examples/) on it. Please, change your urls accordingly.

BelousSofiya: Thank you for feedback. But we have this naming of urls in our profiles: /profiles/ and profiles/<pk>. And Heorgii in last meeting told that we need to write some uniform code. So I tried to write similar to existing urls

profiles/tests/test_activity.py
BelousSofiya: Thank you for feedback. Done.

profiles/tests/test_category.py
BelousSofiya: Thank you for feedback. Done

profiles/views.py
BelousSofiya: Thank you for feedback. Done

profiles/views.py
BelousSofiya: Thank you for feedback. Done

profiles/permissions.py
BelousSofiya: Thank you for feedback.  I think my varient is more readable.

profiles/urls.py
BelousSofiya: Thank you for feedback. Done

profiles/models.py
```diff
@@ -68,12 +68,12 @@ class Profile(models.Model):
 
 class Activity(models.Model):
     activity_id = models.AutoField(primary_key=True)
```
ProDG, line 70: I see that it's not from this PR, but still: if you need such kind of field in your model, it is created automatically by default and should not be declared explicitly.

profiles/permissions.py
ProDG: Actually, the best option (and the option which I strongly recommend) will be implemenatation of narrow-focused permission classes and then combining it in `permission_classes` in view.

`Or`, `And` etc words in names of permission classes are a clear sign of violation of SRP. Thus, it's hard to reuse them etc.

So, in this case, instead of creation `IsAdminOrReadOnly` permission class, 2 permission classes (something like `IsAdminUser` + `RequestMethodIsSafe`) can be created and used as `permission_classes = (RequestMethodIsSafe | IsAdminUser)` in a view.

profiles/urls.py
```diff
@@ -12,6 +13,8 @@
     path('saved-list/<pk>/', SavedCompaniesDestroy.as_view(), name='saved_companies_destroy'),
     path('viewed-list/', ViewedCompanyList.as_view(), name='viewed_company_list'),
     path('regions/', RegionListView.as_view(), name='region-list'),
-    path('category-list/', CategoryList.as_view(), name='category_list'),
-    path('activity-list/', ActivityList.as_view(), name='activity_list'),
+    path('categories/', CategoryList.as_view(), name='category_list'),
+    path('activities/', ActivityList.as_view(), name='activity_list'),
+    path('categories/<pk>', CategoryDetail.as_view(), name='category-detail'),
```
ProDG, line 18: Add trailing `/` (here and in all other similar views).

ProDG (approved):

profiles/models.py
BelousSofiya: Oleksandra and Diana work on this problem in other PR

PR #209, opened 2023-09-17
#206 fix all saved companies filter

Lvyshnevska:
All the filters (is saved, company type, activities) have been redone with django filter. Related to these filters tests have been updated.


profiles/permissions.py
```diff
@@ -13,7 +13,7 @@ def has_object_permission(self, request, view, obj):
 
 class SavedCompaniesListPermission(BasePermission):
     def has_permission(self, request, view):
-        if request.query_params.get("filters") == "is_saved":
+        if request.query_params.get("is_saved"):
```
ProDG, line 16: Not nice, normally permissions should not have any relations with filtering.

I don't mind going with this option (it will be definitely not the most questionable code in this project). But if you have time and inspiration, I'd consider the next options:

1. Just ignore `is_saved` in permission class (this actually will be right approach, because filtering options should not be a reason to return 403). Of course, you'll have to update logic in this filter (in such way, that when unauthorized user tries to use this filter — empty queryset is returned).
2. Instead of using the same endpoint for 'normal' profiles list, and saved profiles list — create two separate endpoints. The one for saved profiles list will have `IsAuthenticated` in permission classes.

Both options are equally OK (each has own advantages/disadvantages, and in different cases I used both of them in real projects) and both are much better than the current approach.

ProDG (approved):

profiles/permissions.py
Lvyshnevska: Permission removed and logic in is_saved filter updated.

Thanks for the recommendation!

Shushunya (approved):

profiles/filters.py
```diff
@@ -0,0 +1,32 @@
...
+            if self.request.user.is_authenticated:
+                queryset = queryset.filter(saved_list__user=self.request.user)
+            else:
+                queryset = queryset.none()
+        return queryset
+
+    def company_filter(self, queryset, name, value):
+        if value:
+            return queryset.filter(is_registered=True)
+        return queryset
```
ProDG, line 22: Sorry for not mentioning it — when using `field_name`, defining a method is redundant. It should work without it.

ProDG: ... and also, when filter name matches field name, you don't need even pass `field_name` argument.

```
is_startup = filters.BooleanFilter()
```
... and that's it. 

It also will be better this way, than a current implementation. As there will be no confusion between filter name and actual field name.

Lvyshnevska: I'll remove redundant methods. Is it obligatory to name filters as field names in model? Or my implementation is option too?

PR #219, opened 2023-09-20
#186connect with swr

ddianaoo:
The SWR was I investigated. In this PR I used it for a GET request, so I suggest using it as per my example.


FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -248,15 +233,19 @@ const GeneralInfo = (props) => {
...
                             name='regions'
-                            options={REGIONS}
+                            options={fetchedRegions}
                             label={LABELS.regions}
-                            updateHandler={onUpdateSelectField}
+                            updateHandler={onUpdateOneSelectField}
                             requredField={false}
-                            value={user.regions}
                             defaultValue="Оберіть"
+                            value={fetchedRegions.find((el) => el.key ===  user.regions)
```
popovycholeg, line 244: It's better to use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) and [optinal chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `fetchedRegions.find((el) => el.key ===  user.regions)?.value ?? ''`.

forum/settings.py
```diff
@@ -65,13 +65,13 @@
 CORS_ALLOW_ALL_ORIGINS = False
 
 CORS_ORIGIN_WHITELIST = [
-    'http://localhost',
+    'http://localhost:3000',
```
popovycholeg, line 68: Could you whitelist `http://178.212.110.52` as well?

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -248,15 +233,19 @@ const GeneralInfo = (props) => {
                             value={user.edrpou}
                             error={edrpouError}
                         />
-                        <MultipleSelectChip
+                        {!isLoading &&
```
popovycholeg, line 236: Should we show the loading state? Do we have any kind of error state?

popovycholeg (approved):
LGTM in general! Left some comments

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -90,13 +64,17 @@ const ERRORS = {
...
+const fetcher = (...args) => fetch(...args).then(res => res.json());
+
 const GeneralInfo = (props) => {
     const [user, setUser] = useState(props.user);
     const [formStateErr, setFormStateErr] = useState(ERRORS);
     const [imageBannerError, setImageBannerError] = useState(null);
     const [imageLogoError, setImageLogoError] = useState(null);
     const [edrpouError, setEdrpouError] = useState(null);
 
+    const { data: fetchedRegions, error, isLoading } = useSWR('http://127.0.0.1:8000/api/regions/', fetcher);
```
popovycholeg, line 76: Could we use `.env` for storing the base URLs?

forum/settings.py
ddianaoo: Our team is currently working on this, so we`ll add it soon.

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
ddianaoo: I decided not to implement this for now because I remember my teammate wanted to complete this task. I only added this condition, because, otherwise I might get an error.

Krymets (approved):

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
ddianaoo: Our team is currently working on this, too We`ll implement it)

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
ddianaoo: Thanks, I added this to a new commit. I'm a beginner in js so didn't know about these ways but glad to see your comments which helped me improve my work)

PR #221, opened 2023-09-20
routeswith slash

BelousSofiya:
Add slash to routes for activity and category, fix routes in tests

### DIFF
--- a/profiles/tests/test_activity.py
+++ b/profiles/tests/test_activity.py
@@ -1,4 +1,4 @@
-from rest_framework.test import APITestCase, APIClient
+from rest_framework.test import APITestCase
 
 from authentication.factories import UserFactory
 from profiles.factories import ActivityFactory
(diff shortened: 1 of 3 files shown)


profiles/tests/test_activity.py
```diff
@@ -18,10 +18,18 @@ def test_get_all_activities_unauthorized(self):
         response = self.client.get("/api/activities/")
         self.assertEqual(200, response.status_code)
 
-    def test_get_all_activities_data(self):
-        response = self.client.get("/api/activities/")
-        self.assertEqual([{'id': 46, 'name': 'sale'}, {'id': 47, 'name': 'transport'}, {'id': 48,
-                            'name': 'education'}, {'id': 49, 'name': 'medicine'}], response.json())
+    def test_get_activity_data_authorized_admin(self):
+        self.client.force_authenticate(self.test_person_is_admin)
+        response = self.client.get(f"/api/activities/{self.sale_activity.id}/")
```
ProDG, line 23: I recommend to use `.format()` with named arguments instead of f-strings.

ProDG: ... in this case.

Reason: properly named placeholders usually give clearer understanding what is in this part of URL, than some value directly put into f-string.

BelousSofiya: Why?

ProDG (approved):

Shushunya (approved):

VovaD199 (approved):

PR #223, opened 2023-09-22
217 connect signup

Shushunya:
Added POST axios request to onSubmit method. Error handling is another task so it is not implemented properly.

### DIFF
--- a/FrontEnd/package.json
+++ b/FrontEnd/package.json
@@ -44,5 +44,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "dotenv": "^16.3.1"
   }
 }
--- a/FrontEnd/package.json
+++ b/FrontEnd/package.json
@@ -17,6 +17,7 @@
     "react-router-dom": "^6.4.1",
     "react-router-hash-link": "^2.4.3",
     "react-scripts": "5.0.1",
+    "swr": "^2.2.2",
     "validator": "^13.11.0",
     "web-vitals": "^2.1.4"
   },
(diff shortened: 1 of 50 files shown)


FrontEnd/package.json
```diff
@@ -17,6 +17,7 @@
     "react-router-dom": "^6.4.1",
     "react-router-hash-link": "^2.4.3",
     "react-scripts": "5.0.1",
+    "swr": "^2.2.2",
```
popovycholeg, line 20: Do you use it?

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
```diff
@@ -61,13 +62,33 @@ export function SignUpFormContentComponent(props) {
...
+        is_startup: isChecked.startup,
+      },
+    };
+    
+    axios({
+      method: 'post',
+      url: `http://localhost:8000/api/auth/users/`,
+      withCredentials: false,
+      data: dataToSend
+    }).then(res => console.log(res.data)).catch(error => console.log(error))
```
popovycholeg, line 89: Should we do smth then after registration? I know that API doesn't return a token, but should we redirect the user to the logic page?

Shushunya: I suppose there's another task to add modal window. But there are some inconsistences so I left it for the next task.

FrontEnd/package.json
Shushunya: Now it's not in my PR. Thanks for comment :)

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
Shushunya: I left ToDos for the future work.

Lvyshnevska (approved):

PR #233, opened 2023-09-25
#205 Refactor Serializers

ddianaoo:
1. First of all, ProfileSerializer was renamed to ProfileListSerializer, and we use it only for view the list of profiles. There I also removed the phone field (sensitive data) and other attributes that don`t need to be shown. 
2. I added ProfileCreateSerializer, because we had some problems with displaying full data of activities and categories.
3. In ProfileDetailSerializer I added the is_saved field because this serializer is used when we want to see some profile and we need to be able to save this company.
4. ProfileOwnerDetailViewSerializer was added with displayed sensitive data. So I added the functionality that will give us an ability to view our company like other peoples see it.
5. Also was added the ProfileOwnerDetailEditSerializer with full data, that the owner can change the info about his company.
6. I removed the redundant serializer - UserTokenCreateSerializer, because we realized that its functionality is useless for us.
7. added a display_region field for comfort usage it in frontend. 
8. And tests were corrected according to my changes


profiles/serializers.py
```diff
@@ -34,17 +31,34 @@ def get_is_saved(self, obj):
 
 
 class ProfileDetailSerializer(serializers.ModelSerializer):
-    activity = ActivitySerializer(many=True, read_only=True)
-    category = CategorySerializer(many=True, read_only=True)
+    is_saved = serializers.SerializerMethodField()
+    email = serializers.ReadOnlyField(source='person.email')
```
Shushunya, line 35: It's a serializer with contacts being hidden, right? Why do we have here `email` field? I thought it supposed to be hidden :)

profiles/serializers.py
```diff
@@ -17,14 +15,13 @@ class Meta:
 
 
 class ProfileSerializer(serializers.ModelSerializer):
```
Shushunya, line 17: Better rename it. I suggest `ProfileListSerializer` to avoid misapprehension in future.

profiles/serializers.py
ddianaoo: thank you) it was fixed in my new commit

profiles/serializers.py
ddianaoo: was renamed. Thanks

GeorgiyII (approved):

Shushunya (approved):

Lvyshnevska (approved):

profiles/serializers.py
```diff
@@ -16,15 +14,14 @@ class Meta:
...
-    activity = ActivitySerializer(many=True, read_only=True)
-    category = CategorySerializer(many=True, read_only=True)
-    person = serializers.PrimaryKeyRelatedField(queryset=CustomUser.objects.all())
+class ProfileListSerializer(serializers.ModelSerializer):
     is_saved = serializers.SerializerMethodField()
 
     class Meta:
         model = Profile
-        fields = "__all__"
+        fields = ('name', 'person', 'is_registered', 'is_startup', 'official_name', 'region', 'common_info', 'address',
```
Shushunya, line 22: Alter returned `activities` and `categories` fields so that they contain their names. Right now we get just lists of numbers (IDs) accordingly and so we'll have to make extra requests on the frontend to simply get those names of activities and categories.

Shushunya (changes requested):

profiles/serializers.py
ddianaoo: fixed)

Shushunya (approved):

Shushunya (approved):

ProDG (approved):

PR #243, opened 2023-09-29
#192 saved company bug

BelousSofiya:
Fix bug with adding user's own company to SavedList
Fix bug to add company to another user's saved list

### DIFF
--- a/administration/serializers.py
+++ b/administration/serializers.py
@@ -1,6 +1,7 @@
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
+from djoser.serializers import UserSerializer
 
 from authentication.models import CustomUser
 from profiles.models import Profile
@@ -9,7 +10,7 @@ from profiles.models import Profile
 User = get_user_model()
 
 
-class AdminUserSerializer(serializers.UserSerializer):
+class AdminUserSerializer(UserSerializer):
 
     phone_number = serializers.SerializerMethodField()
 
@@ -19,7 +20,7 @@ class AdminUserSerializer(serializers.UserSerializer):
         except ObjectDoesNotExist:
             return None
 
-    class Meta(serializers.UserSerializer.Meta):
+    class Meta(UserSerializer.Meta):
         model = User
         fields = (
             "id",
--- a/administration/serializers.py
+++ b/administration/serializers.py
@@ -16,7 +16,7 @@ class AdminUserSerializer(UserSerializer):
 
     def get_phone_number(self, user):
         try:
-            return user.profile.comp_phone_number
+            return user.profile.phone_number
         except ObjectDoesNotExist:
             return None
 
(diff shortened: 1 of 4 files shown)


VovaD199 (approved):

administration/serializers.py
```diff
@@ -19,7 +20,7 @@ def get_phone_number(self, user):
         except ObjectDoesNotExist:
```
ProDG, line 20: I'm not sure you'll get `ObjectDoesNotExist` exception in case there are no `profile` instance for the user. Maybe it will be something about attribute etc.

If you haven't checked behavior here, please do this.

profiles/permissions.py
```diff
@@ -15,3 +16,15 @@ class ReadOnly(BasePermission):
...
                  return True
+
+
+class SelfCompany(BasePermission):
+    def has_permission(self, request, view):
+        user = request.user
+        pk = request.data.get("company_pk")
+        try:
+            profile = Profile.objects.get(id=pk)
+        except:
```
ProDG, line 27: Never EVER use `try  .. except` without limiting it to particular exception. `ObjectDoesNotExist` should be suitable here.

profiles/permissions.py
```diff
@@ -15,3 +16,15 @@ class ReadOnly(BasePermission):
     def has_permission(self, request, view):
         if request.method in SAFE_METHODS:
                  return True
+
+
+class SelfCompany(BasePermission):
```
ProDG, line 21: Please consider better name for this permission. Probably something starting with `Is...`, `Has...` etc.

profiles/views.py
```diff
@@ -20,12 +20,12 @@ class SavedCompaniesCreate(CreateAPIView):
...
     Add a company to the saved list.
     """
-    permission_classes = [IsAuthenticated]
+    permission_classes = [IsAuthenticated, SelfCompany]
     serializer_class = SavedCompanySerializer
     pagination_class = ForumPagination
 
     def post(self, request):
-        user = request.user
+        user = self.request.user
```
ProDG, line 28: Why?

ProDG (approved):

profiles/permissions.py
BelousSofiya: Thank you for feedback. Done.

profiles/permissions.py
BelousSofiya: Thank you for feedback. Done.

profiles/views.py
BelousSofiya: Thank you for feedback. I don't remember why)) Delete 'self'

administration/serializers.py
BelousSofiya: Thank you for feedback. I've just fix bug with import in this file. But I'll tell the author of this function about your advice.

whooaami (approved):

GeorgiyII (approved):

PR #244, opened 2023-09-29
241 add loader

Shushunya:
image


FrontEnd/src/components/loader/Loader.jsx
```diff
@@ -0,0 +1,36 @@
+import css from "./Loader.module.css";
+import { Carousel, ConfigProvider } from "antd";
+
+export default function Loader() {
+  return (
+    <div className={css.container}>
+      <ConfigProvider
```
olehpopovych-amplify, line 7: I think config provide should be moved to app root with rest providers

FrontEnd/src/components/loader/Loader.jsx
```diff
@@ -0,0 +1,36 @@
+import css from "./Loader.module.css";
```
olehpopovych-amplify, line 1: Could we add a loader to the page where we fetch regions?
Also, I would really appreciate it if you add some screenshots to the pr description

Shushunya: Screenshot added to the description of the PR :)

FrontEnd/src/components/loader/Loader.jsx
```diff
@@ -0,0 +1,36 @@
...
+          </div>
+          <div>
+          <span/>
+          </div>
+          <div>
+          <span/>
+          </div>
+          <div>
+          <span/>
+          </div>
```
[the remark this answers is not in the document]
Shushunya: done

Shushunya:
One test on the server side is still failing. But my PR was purely the client side part. Also I've changed 2 tests in authorization so now they also pass.

PR #246, opened 2023-09-29
245 Add 404 Not Found error page

Shushunya:
image

Also changes according to comments about `ErrorPage` from #248


Shushunya:
Client side task. Test on the server side is failing.

PR #248, opened 2023-09-29
Add profile list page

Shushunya:
Added Profile List Page which already sends requests to the server:
- GET request for profiles (all filters considered)
- POST request to add profile to saved

Also pagination implemented. And state of the page depends on whether the user is logged in.

`ErrorPage` and `Loader` temporary in this task but there are two other according PRs (#246 and #244). These pages will be removed once those two PR approved and merged. Pages are identical. 
**[upd: 04/10 custom loader & error page removed and replaced with correcponding components from dev]**

serializer changes will be gone once PR #233 will be merged **[upd: 03/10 serializer changes removed]**

Issue #124, opened 2023-08-21
Error handling


oksanahorishna:
As a user 
I want to see the error messages 
So that I understand the website behavior 

Server errors and client errors should be mapped and displayed to the user in Ukrainian

PR #256, opened 2023-10-08
Sign up error handling

Shushunya:
part of the #124 task [Sign Up: 400 email exists]

2 first commits are duplicated from #255 PR since they are fixing eslint errors after merge conflicts in #248 

after #255 is merged, only 1 file will have changes: `SignUpFormContent`, so please have look only at it

[upd: 09/10 now is up to date state]

### DIFF
--- a/FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
+++ b/FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
@@ -34,6 +34,7 @@ export function SignUpFormContentComponent(props) {
     handleSubmit,
     watch,
     getValues,
+    setError,
     formState: { errors, isValid },
   } = useForm({
     mode: 'all',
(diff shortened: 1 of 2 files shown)


Lvyshnevska (approved):

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
```diff
@@ -83,12 +84,18 @@ export function SignUpFormContentComponent(props) {
     axios({
       method: 'post',
       url: `${process.env.REACT_APP_BASE_API_URL}/api/auth/users/`,
-      withCredentials: false,
-      data: dataToSend
-    }).then(res => console.log(res.data)).catch(error => console.log(error));
-    console.log(process.env.REACT_APP_BASE_API_URL);
+      data: dataToSend,
+    })
+      .then()
```
popovycholeg, line 89: No need to leave empty `then` here

popovycholeg: But should the user somehow be informed that registration is successfull?

popovycholeg (approved):

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
Shushunya: It will be used to trigger notification modal #262 as far as I know. That's why I decided to leave it even though it's empty.

PR #261, opened 2023-10-09
#260 add check password when delete profile

Lvyshnevska:
Functionality of checking password when delete profile added (destroy method overridden). Tests (delete profile) updated due to changes.

### DIFF
--- a/profiles/views.py
+++ b/profiles/views.py
@@ -180,9 +180,7 @@ class ProfileDetail(RetrieveUpdateDestroyAPIView):
         instance = self.get_object()
         user = self.request.user
         password = self.request.data.get("password")
-        if not password:
-            return Response(status=status.HTTP_400_BAD_REQUEST)
-        if not check_password(password, user.password):
+        if not password or not check_password(password, user.password):
             return Response(status=status.HTTP_400_BAD_REQUEST)
         else:
             self.perform_destroy(instance)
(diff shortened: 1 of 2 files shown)


profiles/views.py
```diff
@@ -174,6 +176,16 @@ def get_serializer_class(self):
         else:
             return ProfileOwnerDetailEditSerializer
 
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
```
Shushunya, line 180: Maybe it's better to use `get_object_or_404` here?

Shushunya (approved):

profiles/views.py
```diff
@@ -174,6 +176,16 @@ def get_serializer_class(self):
...
+        user = self.request.user
+        password = self.request.data.get("password")
+        if not password or not check_password(password, user.password):
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
ProDG, line 184: 1. `400 BAD REQUEST` must have a body, with explanation what exactly is wrong with the request. So, I guess this logic (together with password check) should be in serializer validator.
2. Don't, don't do `return Response()`.

ProDG (approved):

profiles/views.py
Lvyshnevska: I wanted to do it in serializer, but there is no delete method in serializers, so I don't know how to implement it. I remember that it's not okay to return Response. But I just override original destroy method and it contains Response.

PR #263, opened 2023-10-10
#add search functionality by makar

BelousSofiya:
Makar made a greate job and I don't want it to diy in that ancient PR. So I'll try to move it to develop. Just backend part

### DIFF
--- a/search/filters.py
+++ b/search/filters.py
@@ -1,5 +1,5 @@
 from django_filters import filters
-from django_filters.rest_framework import FilterSet
+from django_filters.rest_framework import FilterSet, OrderingFilter
 
 
 class CompanyFilter(FilterSet):
--- a/search/filters.py
+++ b/search/filters.py
@@ -3,15 +3,5 @@ from django_filters.rest_framework import FilterSet
 
 
 class CompanyFilter(FilterSet):
-    is_saved = filters.BooleanFilter(method="is_saved_filter")
-    is_registered = filters.BooleanFilter()
-    is_startup = filters.BooleanFilter()
-    activities__name = filters.CharFilter()
-
-    def is_saved_filter(self, queryset, name, value):
-        if value:
-            if self.request.user.is_authenticated:
-                queryset = queryset.filter(saved_list__user=self.request.user)
-            else:
-                queryset = queryset.none()
-        return queryset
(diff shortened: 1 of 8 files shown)


search/views.py
```diff
@@ -0,0 +1,16 @@
...
+from .serializers import CompanySerializers
+
+
+class SearchCompanyView(ListAPIView):
+    serializer_class = CompanySerializers
+
+    def get_queryset(self):
+        search_field = self.request.query_params.get(
+            "search_field", ""
+        ).strip()
```
ProDG, line 13: Don't we use Django Filters?

What if we'll have to do search by multiple fields, + user-controlles ordering?

BelousSofiya: Thank you for feedback. Changed it.

search/filters.py
```diff
@@ -0,0 +1,21 @@
...
+from django_filters.rest_framework import FilterSet
+
+
+class CompanyFilter(FilterSet):
+    name = filters.CharFilter(method="by_name_filter")
+    region = filters.CharFilter(method="by_region_filter")
+
+    def by_name_filter(self, queryset, name, value):
+        if value:
+            name = self.request.query_params.get("name", "").strip()
```
ProDG, line 11:
```
name = value.strip()
```

search/filters.py
```diff
@@ -0,0 +1,21 @@
...
+            return queryset.filter(name__icontains=name).order_by("name")
```
ProDG, line 12: For ordering there are ordering filters.

These filters only have to apply filtering.

search/filters.py
BelousSofiya: Thank you for feedback. Done.

search/filters.py
BelousSofiya: Thank you for feedback. Deleted ordering.

ProDG: In general, in DRF projects, I strongly recommend to be very careful with usage of `.order_by()`. Or you can end up with cases when you need to have ordering filters, and they don't work because ordering is already applied somewhere else.

whooaami (approved):

search/filters.py
BelousSofiya: I've tried. Is this smth that you've meant?

PR #267, opened 2023-10-12
#Improve check password on profile delete

Lvyshnevska:
Check password on profile delete was improved. Password validation moved to serializer.

### DIFF
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -208,8 +208,8 @@ class ProfileOwnerDetailEditSerializer(serializers.ModelSerializer):
         password = data.get("password")
         if not password:
             raise serializers.ValidationError("Password is required")
-        request_user = self.context["request"].user
-        if not request_user.check_password(password):
+        user = self.context["request"].user
+        if not user.check_password(password):
             raise serializers.ValidationError("Invalid password")
         return data
 
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -208,9 +208,6 @@ class ProfileOwnerDetailEditSerializer(serializers.ModelSerializer):
 class ProfileDeleteSerializer(serializers.Serializer):
     password = serializers.CharField(write_only=True, required=True)
 
-    # class Meta:
-    #     fields = ("password",)
-
     def validate(self, data):
         password = data.get("password")
         user = self.context["request"].user
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -204,6 +204,15 @@ class ProfileOwnerDetailEditSerializer(serializers.ModelSerializer):
         )
         read_only_fields = ("person",)
 
+    def validate_for_delete(self, data):
+        password = data.get("password")
+        if not password:
+            raise serializers.ValidationError("Password is required")
+        request_user = self.context["request"].user
+        if not request_user.check_password(password):
+            raise serializers.ValidationError("Invalid password")
+        return data
+
 
 class ProfileSensitiveDataROSerializer(serializers.ModelSerializer):
     email = serializers.ReadOnlyField(source="person.email")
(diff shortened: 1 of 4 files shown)


profiles/serializers.py
```diff
@@ -204,6 +204,15 @@ class Meta:
         )
         read_only_fields = ("person",)
 
+    def validate_for_delete(self, data):
+        password = data.get("password")
+        if not password:
+            raise serializers.ValidationError("Password is required")
```
ProDG, line 210: just write `required=True` in the serializer field, and there will be no need to make it 'manually'.

profiles/views.py
```diff
@@ -175,19 +174,12 @@ def get_serializer_class(self):
...
-            return Response(status=status.HTTP_204_NO_CONTENT)
-
     def perform_destroy(self, instance):
-        instance.is_deleted = True
-        instance.save()
+        request_data = {"password": self.request.data.get("password")}
+        serializer = self.get_serializer(instance)
+        if serializer.validate_for_delete(request_data):
+            instance.is_deleted = True
+            instance.save()
```
ProDG, line 182:
```python
def perform_destroy(self, instance):
    instance.is_deleted = True
    instance.save()
```
↑ woudn't it be enough?

profiles/views.py
```diff
@@ -175,19 +174,12 @@ def get_serializer_class(self):
...
-        else:
-            self.perform_destroy(instance)
-            return Response(status=status.HTTP_204_NO_CONTENT)
-
     def perform_destroy(self, instance):
-        instance.is_deleted = True
-        instance.save()
+        request_data = {"password": self.request.data.get("password")}
+        serializer = self.get_serializer(instance)
+        if serializer.validate_for_delete(request_data):
```
ProDG, line 180: Why do we manually call the validation method?

Write the validation method. Use it on serializer field. DRF will care about the rest.

profiles/serializers.py
Lvyshnevska: Fixed

profiles/views.py
Lvyshnevska: The serializer is not engaging in this case.

profiles/views.py
Lvyshnevska: Fixed

profiles/views.py
```diff
@@ -172,22 +172,16 @@ def get_serializer_class(self):
...
-            return Response(status=status.HTTP_400_BAD_REQUEST)
-        else:
-            self.perform_destroy(instance)
-            return Response(status=status.HTTP_204_NO_CONTENT)
-
     def perform_destroy(self, instance):
-        instance.is_deleted = True
-        instance.save()
+        serializer = self.get_serializer(data=self.request.data)
+        if serializer.is_valid(raise_exception=True):
```
ProDG, line 182: There is no need do put it into `if`. Just call `serializer.is_valid(raise_exception=True)`, if will raise exception and return 400 BAD REQUEST in case of validation issues anyway.

profiles/serializers.py
```diff
@@ -204,6 +204,15 @@ class Meta:
...
+        user = self.context["request"].user
+        if not user.check_password(password):
+            raise serializers.ValidationError("Invalid password")
```
ProDG: This is not exactly correct. There should be a field name, + probably validation error should not be a string, but a list of strings (with one string in the list).

Lvyshnevska: This code is outdated. Here is how code looks like now: 
if not user.check_password(password):
            raise serializers.ValidationError({"password": "Invalid password"})

profiles/views.py
Lvyshnevska: Fixed

profiles/serializers.py
Lvyshnevska: Replaced with field validation.

Shushunya (approved):

PR #268, opened 2023-10-12
Adjust GitHub workflows

Shushunya:
Trying to adjust github workflows so they are triggered properly:
- BE tests & linter only on BE changes
- FE linter only on FE changes

Result:
 No checks triggered if no corresponding changes (no `.py` or from `FrontEnd` files);
image

 `Eslint` is triggered only if changes in `FrontEnd` dir are present;
image

 `Lint` and `Testing BE` are triggered only if changes in `.py` files are present;
image

 All three are triggered if PR contains both BE (`.py` files) and FE (`FrontEnd` dir files) changes.
image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,5 +1,8 @@
 name: Testing BE
-on: pull_request
+on: 
+  pull_request:
+    paths:
+      - '!FrontEnd/**'
 
 env:
   SECRET_KEY: django-insecure-8p_=_(i8$#=t_n26md(d7=gpc%5sss!4-1a0cp(lir3-0x^$8%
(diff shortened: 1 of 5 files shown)


Shushunya:
With no changes in `FrontEnd` or `.py` files no existing checks run.

Shushunya:
Adding changes in `.py` files [only BE] triggers `Test BE` & `Lint` (BE checks). `EsLint` is not triggered.

.github/workflows/tests.yml
```diff
@@ -1,5 +1,8 @@
 name: Testing BE
-on: pull_request
+on: 
+  pull_request:
+    paths:
+      - '**.py'
```
ProDG, line 5: I'd recommend to not skip Python unittests in case there were no changes in Python. Other things can change (environment variables, for example), unittests execution doesn't take too long, so there is a sense to run it on each deployment.

But it's up to you. I'm approving PR.

ProDG (approved):

PR #287, opened 2023-11-02
add FE tests to git actions

BelousSofiya:


.github/workflows/tests_fe.yml
```diff
@@ -0,0 +1,26 @@
+name: Testing FE
+
+on: pull_request
+
+jobs:
+  build_and_test:
+    name: Testing # job name (unique id)
```
popovycholeg, line 7: Since we test FE it's better to mention it in the name like `Frontend testing`

.github/workflows/tests_fe.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest # on which machine to run
+    steps: # list of steps
+      - name: Install NodeJS
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
popovycholeg, line 13: Could we run tests over the newer node version?
https://nodejs.org/en/about/previous-releases

popovycholeg (approved):

.github/workflows/tests_fe.yml
BelousSofiya: For now I named it 'Testing FE', because backend tests in our git actions is named 'Testing BE'

.github/workflows/tests_fe.yml
BelousSofiya: Thank you for feedback. It was very usefull! I've changed node version to 18 both in tests and eslint.

PR #298, opened 2023-11-07
273 adjust viewed company logic

Shushunya:
`ViewedCompany` model changed:
- `user `can be blank & null field (so the requirement for the anonymos users met)
- properties `user_profile_name` and `company_name` added (due to requirements)
- migration added.

`ViewedCompanySerializer` is changed:
- fields for properties added;
- field `user_profile_name` shows info only if the user is not anonymos;
- `create` method added so it creates an instance properly (is `user` is anonymos, the instance doesn't have it);
- `validate` method removed since now the check whether the user is owner is on the permission classes.

`ViewedCompanyAPIView` changes:

- `ViewedCompanyList` renamed to `ViewedCompanyCreate` for now since we do not have any requirements about retrieving data. So for now this endpoint is only for creating loggs;
- `permission_classes` changed so if the request if `POST` and user is not an owner or is anonymos is proceeded;
- `pagination_class` removed since for now it's only a create api;
- `perform_create` added so the api consumes only `company` data from client side. The user id we get from the request.


profiles/models.py
```diff
@@ -118,8 +118,15 @@ class SavedCompany(models.Model):
...
+    user = models.ForeignKey(
+        CustomUser, on_delete=models.CASCADE, blank=True, null=True
+    )
     company = models.ForeignKey(Profile, on_delete=models.CASCADE)
 
-    class Meta:
-        unique_together = (("user", "company"),)
+    @property
+    def user_profile_name(self):
+        return self.user.profile.name
```
ProDG, line 128: This feature (wrapping some stuff from related models into properties) should be used very carefully.

When other developer will write code related to this model, autocompletion will show him that there is such field on the model. The problem is, that it's not a field on a model, its code which retrieves a chain of 2 other relations. 

So, it's 2 requests for DB, which are a bit obfuscated and it's easy to miss the need to make `select_related`/`prefetch_related`.

profiles/permissions.py
```diff
@@ -13,6 +13,17 @@ def has_object_permission(self, request, view, obj):
...
 
 
+class RequestIsCreate(BasePermission):
+    def has_permission(self, request, view):
+        return request.method == "POST"
+
+
+class IsOwner(BasePermission):
+    def has_permission(self, request, view):
+        profile = Profile.objects.get(id=request.data.get("company"))
```
ProDG, line 23: I strongly recommend to avoid any manipulations with payload (stuff which is meant to be handled by _serializers_) in permission classes.

Current user, view kwargs (i.e. parts of URL) — but not POST/PUT/PATCH payload and not GET-params of URL.

profiles/serializers.py
```diff
@@ -237,18 +238,34 @@ class Meta:
...
+
+    def get_company_name(self, obj):
+        return obj.company_name
+
+    def create(self, validated_data):
+        user = validated_data.get("user")
+        profile = Profile.objects.get(id=validated_data.get("company"))
+        if isinstance(user, AnonymousUser):
+            user = None
+        return ViewedCompany.objects.create(company=profile, user=user)
```
ProDG, line 268: ... in general, logic in this view and serializer is way too complicated.

profiles/views.py
```diff
@@ -184,15 +190,16 @@ def perform_destroy(self, instance):
...
-        return ViewedCompany.objects.filter(user=user_id).order_by(
-            "company_id"
+    queryset = ViewedCompany.objects.all()
+    permission_classes = [
+        (RequestIsCreate & (~IsOwner | ~IsAuthenticated)),
+    ]
+
+    def perform_create(self, serializer):
+        serializer.save(
+            user=self.request.user, company=self.request.data.get("company")
```
ProDG, line 202: Don't overuse `self.request.data`. The stuff in `request.data` is handled in a serializer. You shouldn't need to touch this stuff. 

Very similarly with `return Response()`, if you write `relf.request.data.get(blah-blah)` — 99% you're doing something wrong and there is a better way (alongside with some important stuff from DRF to learn).

profiles/permissions.py
Shushunya: Fixed it, thanks.

profiles/views.py
Shushunya: Fixed.

GeorgiyII (approved):

PR #311, opened 2023-11-15
Increase length of fields in Profile model

Lvyshnevska:
Length of specified fields in Profile model increased:
name - to 100 characters;
common_info - to 2000 characters.


whooaami (approved):

profiles/models.py
```diff
@@ -65,7 +65,7 @@ class Profile(models.Model):
     region = models.CharField(
         max_length=128, choices=Region.choices, default=None, null=True
     )
-    common_info = models.CharField(max_length=255, default=None, null=True)
+    common_info = models.CharField(max_length=2000, default=None, null=True)
```
ProDG, line 68: For texts > 255 chars I'd recommend to use `models.TextField(validators=[MaxLengthValidator(2000)], null=True)`.

`CharField` translates into `varchar` PostgreSQL data type, which is fixed length, i.e. each row in the table will take 2k of disk space, even if the field is empty or contains just a couple of chars. `TextField` translates in dynamic-width data type.

Lvyshnevska: Done.
Thanks.

ProDG (approved):

PR #322, opened 2023-11-23
#321 search page bug

BelousSofiya:
I found some problems on our site. 
1) On search page even your own company had a "like"-star. It doesn't work - you can't add your company to the saved-list. But you see this star. I've fixed it
2) On search page didn't work links to pages with detailed information about company. I've fixed it
3) link /admin redirected us to the inbuild django admin, so our custom admin was not available. I've changed it and for now you can see our custom admin by link /customadmin and Django admin by link /admin
I rebuild our site from this branch, so you can see my changes "in nature"


FrontEnd/src/hooks/useProvideAuth.js
```diff
@@ -4,16 +4,19 @@ import axios from 'axios';
...
         {
           headers: {
             Authorization: `Token ${authToken}`,
           },
         }
-      );
+      )
+      .then((res) => res.data);
+      setUserId(userData.id);
       return true;
```
Lvyshnevska, line 20: You should use existing useUser hook instead of adding userData here.

BelousSofiya: Do you think, we need additional request to api/users/me?

Lvyshnevska: useUser hook provides all the information about authorized user. And you can access this data when you need it. 
It is common practice to use this hook. And it's no need to duplicate the logic of getting user id.
And we get that information via request to api/users/me.

BelousSofiya: When I as authorized user put smth to the searchpage and press enter iuseProvideAuth sends the request to api/users/me. So I use it. If I'll use useUser it will send this request again.

Lvyshnevska: So we have duplicated logic in our code)

BelousSofiya: Propositions?

BelousSofiya: I'll change code, but I'm not sure in this solution

Lvyshnevska: We have useUser hook. My proposition is to use it instead of adding userData in useProvideAuth.

popovycholeg: I agree with @Lvyshnevska , inside a `useUser` hook we utilize a `useSWR` which should cache fetched data and prevent second API call. There is no need to call `/api/auth/users/me/` the second time if it was already called and cached in the `useUser`

BelousSofiya: Ok. Done

whooaami (approved):

FrontEnd/src/hooks/useProvideAuth.js
popovycholeg: As far as what I see we return different values from these hooks. But since they use the same API I assume that we can combine them both into one hook. Also, we do not use SWR in `useProvideAuth` that's something that could be fixed

popovycholeg (approved):

popovycholeg:
Approved, but would be happy to resolve tech debt between `useProvideAuth` and `useUser`. Could you create a ticket for it? @BelousSofiya

PR #323, opened 2023-11-24
#305 Active tab highlighting

Lvyshnevska:
The logic of   active tab highlighting has been updated:
- "Про компанію" is initially highlighted for registered company;
- "Стартап" is initially highlighted for startup;
- another tab is highlighted when user switches to it;
- after page reload, highlighting returns to its initial state;
- only tabs for fields with provided information are shown.

Star button removed for profile owner in profile list.

UPD. Menu tabs functionality has been implemented with Context.

### DIFF
--- a/FrontEnd/src/components/ProfileDetail/MainInfo/ProfileDetailNavBar.jsx
+++ b/FrontEnd/src/components/ProfileDetail/MainInfo/ProfileDetailNavBar.jsx
@@ -1,5 +1,7 @@
 import { HashLink } from 'react-router-hash-link';
-import { useLocation  } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useState, useEffect } from 'react';
+import { PropTypes } from 'prop-types';
 import classes from './ProfileDetailNavBar.module.css';
 
 const MENU_LINKS = {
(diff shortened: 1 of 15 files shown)


FrontEnd/src/components/ProfileDetail/MainInfo/ProfileDetailNavBar.jsx
```diff
@@ -10,38 +12,65 @@ const MENU_LINKS = {
...
+  const navigate = useNavigate ();
+  const [activeLink, setActiveLink] = useState('');
+
+  useEffect(() => {
+    if (hash) {
+      setActiveLink(hash.substring(1));
+    } else {
+      setActiveLink(
+        data.is_registered && document.getElementById('about-company') ? 'about-company' :
+        data.is_startup && document.getElementById('startup') ? 'startup' : ''
```
popovycholeg, line 26: This is a very bad practice to rely on DOM. Do we need to use `document.getElementById` here and could we replace it?

Lvyshnevska: Yes, I read about it. But I didn't find another solution. 
Only tabs for  components with provided information have to be shown. And I wanted to check if component is on the page using id of that component. 
Maybe you can recommend some solution? What is better to use here?

Lvyshnevska: Menu tabs functionality has been implemented with Context.

FrontEnd/src/components/ProfileDetail/MainInfo/ProfileDetailNavBar.jsx
```diff
@@ -10,38 +14,66 @@ const MENU_LINKS = {
...
+  const navigate = useNavigate ();
+  const { dataInComponents } = useContext(DataContext);
+  const [activeLink, setActiveLink] = useState('');
+
+  useEffect(() => {
+    if (hash) {
+      setActiveLink(hash.substring(1));
+    } else {
+      setActiveLink(
+        data.is_registered && dataInComponents.includes('about-company') ? 'about-company' :
```
popovycholeg, line 28: I would recommend using an object with keys instead of the array of strings. 
`
{
  'about-company': true
  ...
}
`

FrontEnd/src/components/ProfileDetail/MainInfo/ProfileDetailNavBar.jsx
```diff
@@ -10,38 +14,66 @@ const MENU_LINKS = {
...
-            }
-          />
-        </div>
-      ))}
+  return (
+    <div>
+      <div className={classes['navbar-menu']}>
+        {Object.entries(MENU_LINKS).map(
+          ([link, label]) =>
+            dataInComponents.includes(link) && (
```
popovycholeg, line 43: Since we store links could we provide a better name for the context that describes it? Like `activeTabs` or `activeLinks`. Same for the context provider

popovycholeg (approved):

Shushunya (approved):

FrontEnd/src/components/ProfileDetail/MainInfo/ProfileDetailNavBar.jsx
Lvyshnevska: Done

Issue #339, opened 2023-12-01
[Profile Edit] Ability to select company/startup


Shushunya:
Profile edit part of the #316

After #316 done, implement:
backend: 
- [ ] check that both options can be choosen after update
frontend:
- [ ] change checkboxes so they both can be selected (same logic as in #334)
- [ ] needed tabs are shown (look at #316 points 2-4)

PR #348, opened 2023-12-05
#339 [Profile Edit] Ability to select company and startup

Lvyshnevska:
Profile editing updated:
   - both Company and Startup options could be selected at the same time;
   - appropriate tabs are active depending on selected option; 
   - tooltips are added for disabled tabs;

### DIFF
--- a/FrontEnd/src/components/ProfilePage/ProfilePageComponents/ProfileContent.js
+++ b/FrontEnd/src/components/ProfilePage/ProfilePageComponents/ProfileContent.js
@@ -53,6 +53,17 @@ const FORM_NAMES = [
 ];
 
 const ProfileContent = (props) => {
+
+    const tooltipInnerContentStyles = {
+            display: 'flex',
+            borderRadius: '2px',
+            background: 'var(--main-grey-90, #25292C)',
+            color: 'var(--main-white, #FFF)',
+            fontFeatureSettings: 'calt',
+            fontFamily: 'Inter',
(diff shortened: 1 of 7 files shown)


FrontEnd/src/components/ProfilePage/ProfilePageComponents/ProfileContent.js
```diff
@@ -52,13 +57,44 @@ const ProfileContent = (props) => {
...
+                            title={element.tooltipText}
+                            overlayInnerStyle={{
+                                display: 'flex',
+                                borderRadius: '2px',
+                                background: 'var(--main-grey-90, #25292C)',
+                                color: 'var(--main-white, #FFF)',
+                                fontFeatureSettings: 'calt',
+                                fontFamily: 'Inter',
+                                letterSpacing: '-0.14px',
+                            }}
```
popovycholeg, line 86: Is it possible to move these styles to the CSS?

popovycholeg (approved):

FrontEnd/src/components/ProfilePage/ProfilePageComponents/ProfileContent.js
Lvyshnevska: I've customized here styles of antd Tooltip component. According to the documentation styles of the tooltip inner content could be changed with overlayInnerStyle property, which type is an object. I didn't find the way to move that styles to the CSS and apply then.

Issue #331, opened 2023-11-29
[Profile] Redirect to Edit my profile screen


mehalyna:
US #325

PR #350, opened 2023-12-06
#331[Profile View] Redirect to edit profile

Lvyshnevska:
Redirect to Edit Profile page from Profile View page added for profile owner.

### DIFF
--- a/FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
+++ b/FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
@@ -92,6 +92,13 @@ function TitleInfo({ isAuthorized, data }) {
     />
   );
 
+  const getStarVisibility = () => {
+    if (isAuthorized) {
+        return isSaved ? filledStar : outlinedStar;
+      }
+    return null;
+  };
+
   const CategoryBadges = ({ categories }) => {
     return (
       <>
@@ -151,7 +158,7 @@ function TitleInfo({ isAuthorized, data }) {
               <span className={classNames(classes['title-block__button--text'], {[classes['added_to_saved__button--text']]: isSaved})}>
                 {!isSaved ? 'Додати в збережені' : 'Додано в збережені'}
               </span>
-              {isAuthorized ? (isSaved ? filledStar : outlinedStar) : null}
+              {getStarVisibility()}
             </button>
           )}
           {ownProfile && (
--- a/FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
+++ b/FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
@@ -96,7 +96,6 @@ function TitleInfo({ isAuthorized, data }) {
     if (isAuthorized) {
         return isSaved ? filledStar : outlinedStar;
       }
-    return null;
   };
 
   const CategoryBadges = ({ categories }) => {
--- a/FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
+++ b/FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
@@ -162,13 +162,13 @@ function TitleInfo({ isAuthorized, data }) {
             </button>
           )}
           {ownProfile && (
-            <button
-              type="button"
-              className={`${classes['title-block__button']}`}
+            <a
+              role="link"
+              className={`${classes['title-block__button']} ${classes['title-block__link']}`}
               onClick={navigateToEditProfile}
             >
               <span className={`${classes['title-block__button--text']}`}>Редагувати профіль</span>
-            </button>
+            </a>
           )}
         </>
       ) : null}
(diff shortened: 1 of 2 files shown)


FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
```diff
@@ -133,15 +139,30 @@ function TitleInfo({ isAuthorized, data }) {
...
-        </button>
+      {isAuthorized ? (
+        <>
+          {!ownProfile && (
+            <button
+              onClick={handleClick}
+              type="button"
+              className={`${classes['title-block__button']} ${isSaved && classes['added_to_saved__button']}`}
+            >
+              <span className={`${classes['title-block__button--text']} ${isSaved && classes['added_to_saved__button--text']}`}>
```
popovycholeg, line 150: You might see a redundant `false` class when `isSaved` is falsy. I would recommend using `classnames` or `clsx`.

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
```diff
@@ -133,15 +139,30 @@ function TitleInfo({ isAuthorized, data }) {
...
+                {!isSaved ? 'Додати в збережені' : 'Додано в збережені'}
+              </span>
+              {isAuthorized ? (isSaved ? filledStar : outlinedStar) : null}
```
popovycholeg, line 153: Could we move it into function? double ternary is not the best practice

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
```diff
@@ -133,15 +139,30 @@ function TitleInfo({ isAuthorized, data }) {
...
+            </button>
+          )}
+          {ownProfile && (
+            <button
+              type="button"
+              className={`${classes['title-block__button']}`}
+              onClick={navigateToEditProfile}
```
popovycholeg, line 160: Should we use `a` instead of a `button`? At least we need to add `role="link"` attribute to it

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
```diff
@@ -133,15 +139,30 @@ function TitleInfo({ isAuthorized, data }) {
...
-            <span className={`${classes['title-block__button--text']} ${isSaved && classes['added_to_saved__button--text']}`}>{!isSaved ? 'Додати в збережені' : 'Додано в збережені'}</span>
-          {isAuthorized ? (isSaved ? filledStar : outlinedStar) : null}
-        </button>
+      {isAuthorized ? (
+        <>
+          {!ownProfile && (
+            <button
+              onClick={handleClick}
+              type="button"
+              className={`${classes['title-block__button']} ${isSaved && classes['added_to_saved__button']}`}
```
popovycholeg, line 148: nit: we didn't care about accessibility so far. It might be worse to revisit our app and provide accessible attributes `tabindex`, `role`, `aria-label`. In the context of another story. But for buttons affected in this story, we might provide `tabindex`

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
Lvyshnevska: Done

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
Lvyshnevska: Done. But I doubt I've made it correctly.

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
Lvyshnevska: Done

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
Lvyshnevska: That's not entirely clear to me. I've tried to understand tabIndex usage and found out that there is no necessity to add tabIndex to button element unless we want to change it's default behavior. As button element has a default tabIndex value 0. 

Could you please explain, what should I do here? Should I add tabIndex attribute to the button "Додати в збережені"? What should be the value?

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
```diff
@@ -85,6 +92,13 @@ function TitleInfo({ isAuthorized, data }) {
     />
   );
 
+  const getStarVisibility = () => {
+    if (isAuthorized) {
+        return isSaved ? filledStar : outlinedStar;
+      }
+    return null;
```
popovycholeg, line 99: I think we can remove this return

popovycholeg (approved):

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
popovycholeg: The general recommendation is to use them, but since we haven't added them to any buttons so far you can skip it. We can pickup this work after MVP

FrontEnd/src/components/ProfileDetail/MainInfo/TitleInfo.jsx
Lvyshnevska: Done

Issue #352, opened 2023-12-08
[Sign Up] Fix redirect to modal window after Sign Up


Lvyshnevska:
Redirect to modal window only if sign up succeed

PR #353, opened 2023-12-08
#352 Fix redirect to modal window after Sign Up

Lvyshnevska:
User is redirected to modal sign-up window only in case of successful registration.

### DIFF
--- a/FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
+++ b/FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
@@ -67,7 +67,11 @@ export function SignUpFormContentComponent(props) {
       url: `${process.env.REACT_APP_BASE_API_URL}/api/auth/users/`,
       data: dataToSend,
     })
-      .then()
+      .then((res) => {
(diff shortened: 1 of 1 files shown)


Shushunya (approved):

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
```diff
@@ -67,7 +67,11 @@ export function SignUpFormContentComponent(props) {
       url: `${process.env.REACT_APP_BASE_API_URL}/api/auth/users/`,
       data: dataToSend,
     })
-      .then()
+      .then((res) => {
+        if (res.status === 201) {
```
popovycholeg, line 71: Do we receive another success status? If no we can remove the if

popovycholeg (approved):

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
Lvyshnevska: Done

PR #354, opened 2023-12-08
#174adding images

BelousSofiya:
Uploading banner_image. User can add image to his company and get image of every company. Only BE part. 
image
image

### DIFF
--- a/images/tests/test_banner.py
+++ b/images/tests/test_banner.py
@@ -1,0 +1,110 @@
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+import os
+from io import BytesIO
+from PIL import Image
+
+from authentication.factories import UserFactory
+from profiles.factories import (
+    ProfileStartupFactory,
+    ProfileCompanyFactory,
+)
+
+from utils.dump_response import dump  # noqa
+
+
+class TestBannerChange(APITestCase):
+    @staticmethod
+    def _generate_image(ext, size=(100, 100)):
+        """for mocking png and jpeg files"""
+        file = BytesIO()
+        image = Image.new("RGB", size=size)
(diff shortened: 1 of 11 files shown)


images/views.py
```diff
@@ -0,0 +1,15 @@
...
+    RetrieveUpdateAPIView,
+)
+from rest_framework.parsers import MultiPartParser, FormParser
+
+from profiles.permissions import UserIsProfileOwnerOrReadOnly
+from profiles.models import Profile
+from .serializers import BannerSerializer
+
+
+class BannerChangeAPIView(RetrieveUpdateAPIView):
```
Shushunya, line 11: I think it's better to rename the view. Something like `BannerRetrieveUpdate` sounds good for me.

Shushunya (approved):

images/views.py
BelousSofiya: Thank you for feedback. Done.

images/tests/test_banner.py
```diff
@@ -0,0 +1,129 @@
...
+        self.wrong_image = self._generate_image("png", (7000, 7000))
+
+        self.user = UserFactory(email="test1@test.com")
+        self.company_dnipro = ProfileStartupFactory.create(
+            person=self.user,
+            name="Dnipro",
+            banner_image=f"banners/{self.right_image.name}",
+        )
+
+        self.company_kiev = ProfileCompanyFactory(name="Kyivbud")
```
ProDG, line 40: #KyivNotKiev. :)

ProDG (approved):

images/tests/test_banner.py
BelousSofiya: Thank you for feedback. Done.

Issue #362, opened 2023-12-13
[Edit Profile] Notification edrpou already exists


Lvyshnevska:
Notify user if company with such edrpou already exists, when saving data in General Info tab.

PR #363, opened 2023-12-13
#362 [Edit Profile] Notification edrpou already exists

Lvyshnevska:
The user is notified if company with such edrpou already exists, when attempting to save data in General Info tab.

UPD.: 
ToastContainer removed from components and added to BasicPage. Also custom styles added to it, it's displayed now as per design.  (Design provides styles for error notifications on Login page.) So all the notifications are styled the same way now. 

image

### DIFF
--- a/FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
+++ b/FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
@@ -194,7 +194,7 @@ const GeneralInfo = (props) => {
                     profileMutate(updatedProfileData);
                 } else if (response.status === 400 ) {
                     const errorData = await response.json();
-                    if (errorData.edrpou) {
+                    if (errorData.edrpou[0] === 'profile with this edrpou already exists.') {
                         toast.error('Компанія з таким ЄДРПОУ вже існує');
                     }
                 }
--- a/FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
+++ b/FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
@@ -1,5 +1,6 @@
 import { PropTypes } from 'prop-types';
 import { useState, useEffect } from 'react';
+import { toast, ToastContainer } from 'react-toastify';
 import useSWR from 'swr';
 import { useUser, useProfile } from '../../../hooks/';
 import css from './FormComponents.module.css';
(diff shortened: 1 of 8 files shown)


BelousSofiya (approved):

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -308,6 +312,7 @@ const GeneralInfo = (props) => {
                     </div>
                 </form>
                 : <Loader />}
+                <ToastContainer position="top-right" autoClose={3000} />
```
popovycholeg, line 315: Since we always show toast messages in one place, I think it is worth to add `ToastContainer` to the page template rather than adding it for every component. We also can remove all other references of it

Lvyshnevska: ToastContainer removed from components and added to BasicPage. Also custom styles added to it, it's displayed now as per design. (Design provides styles for error notifications on Login page.) So all the notifications are styled the same way now.

FrontEnd/src/components/basicPage/BasicPage.jsx
```diff
@@ -30,6 +32,23 @@ function BasicPage() {
...
+    font-size: 14px;
+    border-radius: 2px;
+    box-shadow: none;
+    border: 1px solid var(--Dust-Red-2, #FFCCC7);
+    padding: 9px 16px;
+    gap: 10px;
+  }
+  .Toastify__toast-body {
+    padding: 0;
+  }`;
```
popovycholeg, line 50: I would recommend including these styles as simple CSS. No need to add a new package and increase the bundle size (Unpacked Size: 1.65 MB)

Lvyshnevska: I haven't find possibility to add this styles as simple css. That's why I've implemented it this way (as documentation recommends).
https://fkhadra.github.io/react-toastify/how-to-style/

Lvyshnevska: I've styled it in a little bit easier way, I suppose. 
Is it okay?

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -191,8 +192,11 @@ const GeneralInfo = (props) => {
                 if (response.status === 200) {
                     const updatedProfileData = await response.json();
                     profileMutate(updatedProfileData);
-                } else {
-                    console.error('Помилка');
+                } else if (response.status === 400 ) {
+                    const errorData = await response.json();
+                    if (errorData.edrpou[0] === 'profile with this edrpou already exists.') {
+                        toast.error('Компанія з таким ЄДРПОУ вже існує');
+                    }
```
popovycholeg, line 199: I think we want to add else clause with an `unknown error'

popovycholeg (approved):

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
Lvyshnevska: Done

PR #385, opened 2024-01-04
 created a global.css file

ohorodnykostap:
Hello.
In the global.css file, I described the main colors of the web page elements, the font and other changes to normalize the display of tags.
Please review it and I would be happy to receive feedback.


FrontEnd/src/global.css
```diff
@@ -0,0 +1,44 @@
+ @import url('https://fonts.googleapis.com/css2?family=Inter&display=swap');
```
popovycholeg, line 1: I would recommend downloading this font and hosting it using the public/fonts folder to be less dependent on CDN

FrontEnd/src/global.css
```diff
@@ -0,0 +1,44 @@
...
+  --header-bunner-text-color:#000000;
+  --light-text-color: #FFFFFF;
+  --light-button-text-color: #FFFFFF;
+  --delete-text-color: #F15831;
+  --notification-text-color:#F34444;
+  --footer-dark-text-color: #A0ABC0;
+
+  /* button color*/
+  --main-button-color:#1F9A7C;
+  --light-button-color:#FFFFFF;
```
popovycholeg, line 31: nit: space before hex value

FrontEnd/src/global.css
```diff
@@ -0,0 +1,44 @@
...
+  
+  scroll-behavior: smooth;
+  font-family: 'Inter', sans-serif;
+}
+
+a {
+  text-decoration: none;
+}
+
+.img {
```
popovycholeg, line 41: Do you want to apply this style to all `img` tags or classes?

FrontEnd/src/global.css
popovycholeg, line 1: Do we need to import the `global.css` file to the root of our app?

ohorodnykostap: fixed, thanks!

FrontEnd/src/global.css
ohorodnykostap: corrected.

FrontEnd/src/global.css
ohorodnykostap: It should be for all img, but I accidentally put a period. I also had to comment on this part because if you connect global.css, some imgs float.

FrontEnd/src/global.css
ohorodnykostap: Yes, at least that's how I understood the task.

FrontEnd/src/global.css
popovycholeg: Do we need this style at all? I think we can remove it

FrontEnd/src/global.css
popovycholeg: I ask it since I do not see import `global.css` in the PR.

popovycholeg: I ask it since I do not see import `global.css` in the PR. Is the PR ready for review?

FrontEnd/src/global.css
ohorodnykostap: It is really better to delete it.

FrontEnd/src/global.css
ohorodnykostap: Previously, this file was not used in our project and each of the developers wrote styles in their own way. After we add this file, we can use it in work. We can import it into index.css or better into index.js.

popovycholeg: Yes, that is the main point of having `global.css`. I still do not see the import of it in `App.js`

ohorodnykostap: Done.

Lvyshnevska (approved):

popovycholeg (approved):

PR #388, opened 2024-01-06
#221 new members list formation be

BelousSofiya:
Add created_at, updated_at, completeness fields to the Profile model. Add function to migrations to add created_at and updated_at fields to existing profiles.
Add adding points for completed fields banner_image(100), logo_image(1), activities(1), categories(1), region(1)
Add ordering by completeness and created_at in descending order to api/profiles/
Add cleaning commands to django_cd_dev.yml (hello merge conflict with Olexandra's PR)))
Add tests. I have to notice that profile factory creats profile with completed field region. That's why in tests when I update 1 field profile has completeness 2.


profiles/migrations/0007_profile_completeness.py
```diff
@@ -0,0 +1,18 @@
...
+        ('profiles', '0006_profile_created_at_profile_updated_at'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='profile',
+            name='completeness',
+            field=models.SmallIntegerField(default=0),
+        ),
+    ]
```
GeorgiyII, line 18: This action will create the `completeness` with value 0 for all existing profiles.
I think it would be better to add calculation logic in migrations.
Because now all existing profiles on server will have 0 `completeness` until the profile is updated.

https://stackoverflow.com/questions/70588917/django-migrations-calculate-new-fields-value-based-on-old-fields-before-deletin
https://docs.djangoproject.com/en/5.0/ref/migration-operations/

profiles/views.py
```diff
@@ -97,8 +98,12 @@ class ProfileList(ListCreateAPIView):
 
     permission_classes = [IsAuthenticatedOrReadOnly]
     pagination_class = ForumPagination
-    filter_backends = [django_filters.rest_framework.DjangoFilterBackend]
+    filter_backends = [
+        django_filters.rest_framework.DjangoFilterBackend,
+        filters.OrderingFilter,
+    ]
     filterset_class = ProfileFilter
+    ordering = ["-completeness", "-created_at"]
```
GeorgiyII, line 106: This logic with ordering by completeness and created_at in this realization will be applied to all requests to this endpoint.
But if i remember correctly BAs told that we should leave companies list as it is without changes.

Please add ability to send query params from UI client to backend to have an ability to manage what type of sorting, filtering and ordering client wants to have.

profiles/migrations/0007_profile_completeness.py
BelousSofiya: Thank you for feedback. Done

Lvyshnevska (approved):

profiles/migrations/0007_profile_completeness.py
```diff
@@ -0,0 +1,37 @@
...
+                instance.completeness += 100
+            if instance.logo_image:
+                instance.completeness += 1
+            if instance.region:
+                instance.completeness += 1
+            if Activity.objects.all().filter(profile=instance.id):
+                instance.completeness += 1
+            if Category.objects.all().filter(profile=instance.id):
+                instance.completeness += 1
+            instance.save()
```
GeorgiyII, line 28: You already have two places with this code. It would be better to move this logic into function/module and use it here and in profile view

GeorgiyII (approved):

GeorgiyII (approved):

Issue #327, opened 2023-11-24
[Edit Profile] As a registered user, I want to be notified if there are not saved changes when I am leaving the tab, so that I do not lose the information


mehalyna:
#120

PR #391, opened 2024-01-15
#327 [Edit Profile] Notification about unsaved changes

Lvyshnevska:
Notification, that there are unsaved changes on Edit Profile page, if user is leaving/closing the tab, has been added:
- DirtyFormContext has been created
- formIsDirty state for tracking changes in forms has been added
- current field values are compared with default ones (that has been retrieved from db)
- beforeunload event handler has been added
- modal window component WarnUnsavedDataModal has been created
- useBlocker hook has been added to block user from navigating, when formIsDirty set to true
- createBrowserRouter has been added in App.js so that useBlocker could be used
- react-router-dom has been upgraded to latest version to use stable useBlocker hook
- swr key has been updated in useUser and useProfile hooks (to fetch correct data)
- logout functionality has been updated to handle notification about unsaved changes


FrontEnd/src/components/HeaderFooter/header/navbar/Logout.jsx
```diff
@@ -1,23 +1,27 @@
...
-    <button onClick={onClick} type="submit">
-      Вихід
-    </button>
-  );
+  useEffect(() => {
+    const performLogout = async () => {
+      if (auth.isAuth) {
+        try {
+          await axios.post(`${process.env.REACT_APP_BASE_API_URL}/api/auth/token/logout`);
+          await auth.logout();
```
popovycholeg, line 15: Could we move POST request into `auth.logout`?

FrontEnd/src/components/ProfilePage/FormComponents/AdditionalInfo.js
```diff
@@ -14,6 +16,28 @@ const AdditionalInfo = (props) => {
     const { profile: mainProfile, mutate: profileMutate } = useProfile();
     const [profile, setProfile] = useState(props.profile);
     const [foundationYearError, setFoundationYearError] = useState(null);
+    const { setFormIsDirty } = useContext(DirtyFormContext);
+
+    // TODO: update default values as new fields added
```
popovycholeg, line 21: Do we need this todo?

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -65,6 +67,67 @@ const GeneralInfo = (props) => {
...
+                    return;
+                }
+          } else if (defaultValues[key] !== profile[key]) {
+                if (defaultValues[key] === null && profile[key] === '') {
+                    return;
+                }
+                isDirty = true;
+                return;
+          }
+        });
```
popovycholeg, line 122: I would recommend creating a shared util function that will make this check. When we need to make a specific check like for `categories` we may add `type` property to `defaultValues` for running the specific check.

FrontEnd/src/components/ProfilePage/FormComponents/ProductServiceInfo.js
```diff
@@ -16,6 +18,32 @@ const ProductServiceInfo = (props) => {
     const { user } = useUser();
     const { profile: mainProfile, mutate: profileMutate } = useProfile();
     const [profile, setProfile] = useState(props.profile);
+    const { setFormIsDirty } = useContext(DirtyFormContext);
+
+    // TODO: update default values as new fields added
```
popovycholeg, line 23: Do we need this todo on all files?

FrontEnd/src/components/ProfilePage/FormComponents/WarnUnsavedDataModal.jsx
```diff
@@ -0,0 +1,37 @@
...
+            <div className={classes['modal-header']}>
+                <p className={classes['modal-header--text']}>
+                    Збереження введених даних
+                </p>
+                <button onClick={onCancel} className={classes['modal-header--close-icon']}>
+                    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+                        <path d="M8.9257 7.99916L13.6132 2.41166C13.6918 2.31881 13.6257 2.17773 13.5043 2.17773H12.0793C11.9953 2.17773 11.915 2.21523 11.8596 2.27952L7.99356 6.88845L4.12749 2.27952C4.07392 2.21523 3.99356 2.17773 3.90784 2.17773H2.48284C2.36142 2.17773 2.29534 2.31881 2.37392 2.41166L7.06142 7.99916L2.37392 13.5867C2.35631 13.6074 2.34502 13.6327 2.34138 13.6596C2.33774 13.6865 2.3419 13.7139 2.35337 13.7386C2.36484 13.7632 2.38313 13.784 2.40608 13.7985C2.42903 13.8131 2.45568 13.8207 2.48284 13.8206H3.90784C3.99177 13.8206 4.07213 13.7831 4.12749 13.7188L7.99356 9.10988L11.8596 13.7188C11.9132 13.7831 11.9936 13.8206 12.0793 13.8206H13.5043C13.6257 13.8206 13.6918 13.6795 13.6132 13.5867L8.9257 7.99916Z" fill="black" fillOpacity="0.45"/>
+                    </svg>
```
popovycholeg, line 14: I would recommend to move it as a component instead of pasting it as

Shushunya (approved):

FrontEnd/src/components/HeaderFooter/header/navbar/Logout.jsx
Lvyshnevska: I'll try to do it as a part of task on updating logout functionality ( #394 in progress now). This post request is not only in Logout component, but in DeleteProfileModal too.

FrontEnd/src/components/ProfilePage/FormComponents/AdditionalInfo.js
Lvyshnevska: We haven't final profile model with all the fields still. New fields will be added soon or later. I think it's okay to leave this comment as a reminder that default values should be updated then.

FrontEnd/src/components/ProfilePage/FormComponents/ProductServiceInfo.js
Lvyshnevska: We haven't final profile model with all the fields still. New fields will be added soon or later. I think it's okay to leave this comment as a reminder that default values should be updated then.

FrontEnd/src/components/ProfilePage/FormComponents/WarnUnsavedDataModal.jsx
Lvyshnevska: I've moved it to svg files in public folder and added it here with the img tag.

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
Lvyshnevska: Done. checkFormIsDirty util function created.

PR #398, opened 2024-01-28
#326 new members block fe

BelousSofiya:
Create common CompanyCard component for main and search pages. 
Fix FE tests for Search, SearchResults, CompanyCard, add test for MainCompanies.
We decided to remove changes for saved-companies be from this PR.
Unauthorized
image
Authorized
image


Lvyshnevska:
Just first glance at PR (I've pulled the branch and looked through).

Design of cards doesn't correspond to design on Figma. Company cards have different height, logos are placed on different height, company categories are not shown properly. 
image


The functionality of adding company to saved is broken on Profile page. An error occurs when trying to add or remove company.
image


Do we really need to change CreateAPIView view to ListCreateAPIView and add user's id to payload? We don't have such requierements for admin functionality - to get list of all saved companies. And previously we did the opposite changes - ListCreateAPIView to CreateAPIView.

I would highly recommend to request @ProDG  or @GeorgiyII review for backend part of PR.

profiles/permissions.py
```diff
@@ -26,10 +26,18 @@ def has_permission(self, request, view):
         return request.user == view._profile.person
 
 
+class OnlyRequestUserOrAdmin(BasePermission):
```
GeorgiyII, line 29: I don't see the Admin checking here, maybe it would be better to change name of this permission class

I see the logic which checks if method in safe methods and if not then checks if user try to get own user info. So it would be better to name this like OwnerOrReadOnly

BelousSofiya: if request.method in SAFE_METHODS and user.is_staff

user.is_staff is admin user

profiles/permissions.py
```diff
@@ -26,10 +26,18 @@ def has_permission(self, request, view):
...
+    def has_permission(self, request, view):
+        user = request.user
+        if request.method in SAFE_METHODS and user.is_staff:
+            return True
+        return str(request.user.id) == str(request.data.get("user"))
```
ProDG, line 34: Permissions MUST NOT deal with `request.data`.

BelousSofiya: Why?

ProDG: I don't insist on rework of this issue, but strongly recommend to consider it in future.

Ideally, `request.data` should be dealt with only through serializers. Doing `if request.data.get(...)` is pretty close to doing `return Response()`. It works, nobody will die, but is not RESTful. 

Permissions define if user has access to some _specific endpoint URL_. It should not depend on the content of payload, filters (query parameters) etc.

BelousSofiya: And how can we check if user.is_staff or obj.person == request.user (for example)?

profiles/permissions.py
```diff
@@ -26,10 +26,18 @@ def has_permission(self, request, view):
...
+
+
 class IsOwnCompany(BasePermission):
     def has_permission(self, request, view):
         user = request.user
-        pk = request.data.get("company_pk")
+        pk = request.data.get("company")
```
ProDG, line 40: The same issue as above.

I'd recommend to:
1. Put company ID in URL.
2. Extract company in a cached property (`@cached_property` decorator) in a view. If it's detailed view, `get_object()` can be overridden to use value from this property too.
3. In permission class, use this property `view._company`.

Thus, permission class will not have to deal with `request.data`.

profiles/tests/test_saved_company.py
```diff
@@ -29,20 +27,22 @@ def test_add_company_to_saved_authenticated(self):
...
             data={
-                "company_pk": "{profile_id}".format(profile_id=self.profile.id)
+                "user": self.user.id,
+                "company": "{profile_id}".format(profile_id=self.profile.id),
             },
         )
-        self.assertEqual(status.HTTP_200_OK, response.status_code)
-        company_added_info = response.data["Company added"]
-        self.assertEqual(company_added_info["company"], self.profile.id)
+        self.assertEqual(201, response.status_code)
```
ProDG, line 34: `self.assertEqual(status.HTTP_201_CREATED, response.status_code)`

(`status` should be imported somewhere from `rest_framework`).

BelousSofiya: You mean to return 'self.assertEqual(status.HTTP_200_OK, response.status_code)'?

profiles/permissions.py
Shushunya: This kind of check is already implemented if we look at `ViewedCompany` view and permissions.

profiles/permissions.py
GeorgiyII: or OwnerOrAdmin

profiles/tests/test_saved_company.py
ProDG: @BelousSofiya in this case you have status `201`, not `200`. So I guess my suggestion above was correct.

The main point is to use `status` constants instead of just 'magic numbers' with response codes.

profiles/permissions.py
ProDG: > And how can we check if request.user.is_staff or obj.person == request.user (for example)? What is safe method for this?

If you want to check request user against some object — I recommend including the object ID somewhere in URL, retrieving the object in view (through cached property), and then working with this object.

ProDG: @BelousSofiya OK, here is one more point. 

I'm not sure if it work out as a viable exploit of your code in this case, but I believe you'll at least feel the danger. :)

So, in this permission class, you just dive into content of payload. Which is _not validated_ (as at this point it wasn't processed by serializer). You pick some part of this payload, check if it matches current user, and if it does — permission class says 'OK, you can do this action'. Nice and easy, what can go wrong?

But imagine that a bad actor does the action on an object which doesn't actually belongs to him/her. But nothing stops the bad actor to form any payload he wants. So, he puts his ID in `user` field, and this permission class allows him to perform requested action. Funny, isn't it? :)

Now compare this with the approach I suggest. Object ID is part of URL. Object _is retrieved from DB_, and user is compared to the real object's content. 

Btw, object retrieval can be put into something like:

```python
@cached_property
def _my_object(self):
    return get_object_or_404(
        ObjectModel.objects.visible_to(user=self.request.user),
        pk=self.kwargs['object_pk']
    )
```

So, with properly implemented `.visible_to()` manager, code is clean and guarantees immediate returning 404 if someone tries to access the object not available for him.

ProDG: ... in addition to `.visible_to()` we can also have `.editable_by()` etc.

profiles/permissions.py
BelousSofiya: url /saved-list - for create-list functionality. 
url /saved-list/<pk> for delete functionality.
To put id company (not saved-company object id!) to the url with create functionality doesn't seem ok for rest

profiles/permissions.py
BelousSofiya: We have permission for this. User can put not his id in request, but it wouldn't work. But ok, I can delete user.id from request.

ProDG (approved):
OK, let's merge this PR.

It looks like the issue goes far outside of context of this PR, I'll have to review code in the whole and prepare some list of tasks to address issues like this.

profiles/permissions.py
BelousSofiya: Thank you for feedback. Done.

profiles/permissions.py
BelousSofiya: Thank you for feedback. I didn't change permissions a lot (maybe we'll create a task for this), but I've deleted request.data from permissions and user.id from request

profiles/tests/test_saved_company.py
BelousSofiya: Thank you for feedback. Done

profiles/permissions.py
```diff
@@ -26,12 +26,20 @@ def has_permission(self, request, view):
...
-        except ObjectDoesNotExist:
+        if request.method in SAFE_METHODS:
+            return user.is_staff
+        return True
+
+
+class IsOwnCompany(BasePermission):
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
             return True
```
ProDG, line 40: I'm approving this PR. But if would be really nice to follow SRP. Ideally, permission should check ONE thing. The class is called `IsOwnCompany` — so it shouldn't care about request method, it should only check if the company is linked to the current user.

And then these permission classes could be combined in the way we need in `permission_classes` view property.

ProDG (approved):

BelousSofiya:
Done

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+import { useUser } from '../../hooks';
+import PropTypes from 'prop-types';
+import { Tooltip, Badge } from 'antd';
+
+export default function CompanyCard({ data, isAuthorized }) {
+  CompanyCard.propTypes = {
+    data: PropTypes.object,
+    isAuthorized: PropTypes.bool,
+  };
+  const [isSaved, setIsSaved] = useState(data && data.is_saved);
```
popovycholeg, line 16: it's better to pass a callback as a prop to update `data` prop to follow the single responsibility principle

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+      name: data?.name,
+      region: data?.region,
+      categories: data?.categories,
+      activities: data?.activities,
+      logo: data?.logo_image,
+      banner: data?.banner_image,
+      founded: data?.founded,
+      isSaved: data?.is_saved,
+    };
+  }, [data]);
```
popovycholeg, line 31: is there any reason why we can use `data` directly? Does it make sense to provide better naming for `data`  like `profileData` or `profile`

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const { user } = useUser();
+  const ownProfile = user && user.id === profile.personId;
+
+  async function sendRequest(url, { arg: data }) {
+    const authToken = localStorage.getItem('Token');
```
popovycholeg, line 37: FYI: I asked Roman to return token from `useProvideAuth` when `isAuth` `true`. In the future it's better to use token from the hook.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+    return fetch(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Token ${authToken}`,
+      },
+      body: JSON.stringify(data),
+    }).then();
```
popovycholeg, line 45: No need for empty `then`

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+  );
+
+  const handleClick = async () => {
+    try {
+      await trigger(
+        { company_pk: profile.id },
+        { optimisticData: () => setIsSaved(!isSaved) }
+      );
+      mutate(
+        (key) => typeof key === 'string' && key.startsWith('/api/profiles/'),
```
popovycholeg, line 60: Should we the exact api url to update it? I think `/api/profiles/` might cause mutation for a few keys.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+              <div className={styles['company-card__badges_block']}>
+                <CategoryBadges categories={profile.categories.slice(0, 3)} />
+              </div>
+            </Tooltip>
+
+            {isAuthorized && !ownProfile
+              ? isSaved
+                ? filledStar
+                : outlinedStar
+              : null}
```
popovycholeg, line 186: its better to avoid tripple ternary. You can move it to a separate function

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+                key={category.id}
+                size="medium"
+                count={category.name.toUpperCase()}
+                style={{
+                  backgroundColor: '#1F9A7C',
+                  fontWeight: 600,
+                  fontFamily: 'Inter',
+                  fontSize: 10,
+                  margin: 5,
+                }}
```
popovycholeg, line 101: Not a good practice to store inline styles. Could we achieve it with css? If not, we can at least move styles to constant or incapsulate it as a new component if its going to be reusable

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+              />
+            ))
+          : ''}
```
popovycholeg, line 104: Its better to return `null` since an empty string will create an empty text node
https://stackoverflow.com/a/60206733

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+              src={`${process.env.REACT_APP_PUBLIC_URL}/svg/profile-view-image-empty.svg`}
+              alt={profile.name}
+            />
+          )}
+        </div>
+        <div className={styles['company-card__text-block']}>
+          <Tooltip
+            title={
+              profile.activities &&
+              profile.activities.map((element) => element.name).join(', ')
```
popovycholeg, line 131: nit: Could we move it to constant?

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+            }
+            placement="bottom"
+            pointAtCenter={true}
+          >
+            <div className={styles['company-card__category-text']}>
+              {profile.activities &&
+                profile.activities
+                  .slice(0, 3)
```
popovycholeg, line 139: Why 3? Should we show the rest elements in the tooltip? Could we avoid using magic numbers and provide meaningful name for it

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+            <Tooltip
+              title={
+                profile.categories &&
+                profile.categories.map((element) => element.name).join(', ')
+              }
+              placement="bottom"
+              pointAtCenter={true}
+            >
+              <div className={styles['company-card__badges_block']}>
+                <CategoryBadges categories={profile.categories.slice(0, 3)} />
```
popovycholeg, line 178: Wondering why we pass only 3 items? Should we show the rest items withing the tooltip?

FrontEnd/src/components/CompanyCard/CompanyCard.module.css
```diff
@@ -2,6 +2,7 @@
 
 .company-card {
   width: 360px;
+  height: 415px;
```
popovycholeg, line 5: Setting height usually causes many bugs with creating responsive sites. You may use min or max height depending on what you need

FrontEnd/src/components/CompanyCard/CompanyCard.module.css
```diff
@@ -46,6 +52,8 @@
 }
 
 .company-card__text-block {
+  /* width: 360; */
```
popovycholeg, line 55: Do we need to keep it commented?

FrontEnd/src/components/CompanyCard/CompanyCard.module.css
```diff
@@ -46,6 +52,8 @@
...
+  height: 155px;
```
popovycholeg, line 56: Same as above

FrontEnd/src/components/SearchPage/Search.jsx
```diff
@@ -10,53 +8,52 @@ import link_to_left from './img/link_to_left.svg';
...
-    const data = await fetcher(url);
-    setSearchResults(data);
-    setSearchPerformed(true);
-    setError(null);
-  }
+  const fetcher = async (url) => {
+    await axios.get(url, headers).then((res) => {
+      setSearchResults(res.data.results);
+      return res.data.results;
+    });
```
popovycholeg, line 44: Its better not to mix new `async/await` API with `then`. Could you smth one? I prefer `async/await` since its more readable

FrontEnd/src/components/SearchPage/Search.jsx
```diff
@@ -10,53 +8,52 @@ import link_to_left from './img/link_to_left.svg';
...
     if (searchTerm) {
-      try {
-        trigger();
-      } catch (error) {
-        console.error(error);
-      }
+      setSearchPerformed(true);
     }
-  }, [searchTerm, servedAddress, searchUrl]);
+  }, [searchTerm, servedAddress, searchUrl, authToken, companylist]);
```
popovycholeg, line 56: `authToken` should be wrapped in useMemo and have isAuth as a dependency to work properly here. Ideally to do it within `useProvideAuth`. I think Roman is working on it

FrontEnd/src/components/SearchPage/Search.jsx
```diff
@@ -97,21 +95,21 @@ export function Search({ isAuthorized, userData }) {
...
-              <span className={styles['.search_result_error']}>
-                {searchTerm}
-              </span>{' '}
-              не було виявлено на даний момент
-            </p>
-             <br />
-          </div>
+          ) : (
+            <div>
+              <br />
```
popovycholeg, line 103: Do we need `br` here? It seems that we use it for margin here. 
Check

FrontEnd/src/components/landing-page/companies/Companies.jsx
```diff
@@ -0,0 +1,66 @@
...
+import { useState, useEffect } from 'react';
+import styles from './Companies.module.css';
+import CompanyCard from '../../CompanyCard/CompanyCard';
+import PropTypes from 'prop-types';
+import useSWR from 'swr';
+
+const MainCompanies = ({ isAuthorized }) => {
+  MainCompanies.propTypes = {
+    isAuthorized: PropTypes.bool,
+  };
```
popovycholeg, line 11: nit: Usually `propTypes` are placed outside and after component at the end of the file

FrontEnd/src/components/landing-page/companies/Companies.module.css
```diff
@@ -0,0 +1,47 @@
+.new-companies-main {
+  width: 1512px;
+  height: 1106px;
```
popovycholeg, line 3: It is better not to hardcoding height but to use min or max height

FrontEnd/src/tests/CompanyCard.test.js
```diff
@@ -1,131 +1,90 @@
...
-      );
-      expect(screen.getByTestId('star')).toBeInTheDocument();
-      expect(axios.post).toBeenCalledWith(
-        'http://localhost:8000/api/saved-list/',
-        { company_pk: 1 }
-      );
-      expect(axios.get).toHaveBeenCalledWith(
-        'http://localhost:8000/api/profiles/?is_saved=True'
-      );
-    };
```
popovycholeg, line 79: Do we not make post-requests anymore when add or remove start?

FrontEnd/src/tests/MainCompanies.test.js
```diff
@@ -0,0 +1,113 @@
...
+    service_info: null,
+    address: null,
+    banner_image: null,
+  },
+];
+
+describe('MainCompanies component unit tests', () => {
+  test('renders new members results', () => {
+    jest.mock('axios');
+    const axios = require('axios');
```
popovycholeg, line 91: Could we make it in `beforeEach` method?

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: It's in the Lilia's PR. She worked on it. She's waiting for your review. We'll be able change it on all pages after her merge

FrontEnd/src/tests/MainCompanies.test.js
BelousSofiya: Thank you for feedback. Done

FrontEnd/src/components/landing-page/companies/Companies.module.css
BelousSofiya: Our site isn't responsible for now. And we don't have design to implement styles for responsible components. We are going to work on it after having the design

FrontEnd/src/components/landing-page/companies/Companies.jsx
BelousSofiya: Thank you for feedback. Done

FrontEnd/src/components/SearchPage/Search.jsx
BelousSofiya: Thank you for feedback. Done

FrontEnd/src/components/CompanyCard/CompanyCard.module.css
BelousSofiya: Thank you for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.module.css
BelousSofiya: Our site isn't responsible for now. And we don't have design to implement styles for responsible components. We are going to work on it after having the design

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: We show all items in tooltip. Done

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -0,0 +1,209 @@
...
+            </div>
+          </div>
+          <Tooltip
+            title={profile.region}
+            placement="bottom"
+            pointAtCenter={true}
+          >
+            <div className={styles['company-card__region-text']}>
+              {profile.region
+                ? profile.region.length < 35
```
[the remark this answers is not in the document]
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Thank your for feedback. Done.

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
BelousSofiya: Can you explain what do you mean?

FrontEnd/src/components/SearchPage/Search.jsx
BelousSofiya: Can you explain what do you mean?

FrontEnd/src/components/landing-page/companies/Companies.module.css
popovycholeg: I think that it will work the same way w/o hardcoded height. I know but hardcoded height could introduce unexpected behaviour in future

FrontEnd/src/components/CompanyCard/CompanyCard.module.css
BelousSofiya: Thank your for feedback. Done

FrontEnd/src/components/landing-page/companies/Companies.module.css
BelousSofiya: Thank your for feedback. Done

popovycholeg (approved):

PR #399, opened 2024-02-02
#394 [Authorization] Correct logout when token expired

Lvyshnevska:
Authorization updated for correct logout when token expired/401 response status received: 
- AuthProvider was created in useAuth hook (with all the logic of useProvideAuth);
- useUser logic was moved to useAuth;
- useProvideAuth and useUser hooks was removed;
- SWRConfig for global error handling was added (for errors with 401 status);
- Storing of isAuth state was added to local storage, isAuth initial state adjusted (it allows to get correct state when page reloads);
- It was recommended in previous PR to move axios post request from Logout component to logout function. But it wasn't managed. It causes an infinite loop of requests in some cases. So it wasn't changed.

UPD.: 
- fetch requests replaced with axios


BelousSofiya (approved):

FrontEnd/src/components/basicPage/BasicPage.jsx
```diff
@@ -69,29 +67,41 @@ function BasicPage() {
         },
       }}
     >
-      <Header isAuthorized={auth.isAuth} />
+      <SWRConfig value={{
+        onError: (error) => {
+          console.error(error);
+          if (error.status === 401) {
```
popovycholeg, line 73: nit: to avoid magic numbers you might consider using https://www.npmjs.com/package/http-status-codes

FrontEnd/src/hooks/useAuth.js
```diff
@@ -1,6 +1,108 @@
 import { useContext } from 'react';
+import { useEffect, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
+import useSWR from 'swr';
+import axios from 'axios';
 import { AuthContext } from '../context';
 
-export default function useAuth() {
+export function AuthProvider ({ children }) {
+  const [isAuth, setIsAuth] = useState(!!JSON.parse(localStorage.getItem('isAuth')));
```
popovycholeg, line 9: Could we get this data from the SWR?

FrontEnd/src/hooks/useAuth.js
```diff
@@ -1,6 +1,108 @@
...
+      ? [`${process.env.REACT_APP_BASE_API_URL}/api/auth/users/me/`, authToken]
+      : null,
+    ([url, authToken]) =>
+      fetch(url, {
+        method: 'GET',
+        headers: {
+          Authorization: `Token ${authToken}`,
+        },
+      })
+      .then((res) => {
```
popovycholeg, line 26: I think we should set isAuthTrue here when we validate our token. It could be expired

FrontEnd/src/hooks/useAuth.js
```diff
@@ -1,6 +1,108 @@
...
+        console.error(error);
+      }),
+    { revalidateOnFocus: false }
+  );
+
+  const login = (authToken) => {
+    localStorage.setItem('Token', authToken);
+    localStorage.setItem('isAuth', true);
+    axios.defaults.headers.common['Authorization'] = `Token ${authToken}`;
+    setIsAuth(true);
```
popovycholeg, line 52: Could we also set token as state here

FrontEnd/src/hooks/useAuth.js
```diff
@@ -1,6 +1,108 @@
 import { useContext } from 'react';
```
popovycholeg, line 1: Seems that it is not a react hook anymore. Its rather context provider. Could we use appropriate file name than

popovycholeg (approved):
With some comments

FrontEnd/src/hooks/useAuth.js
Lvyshnevska: In case the token has expired, user will be immediately logged out since any request returns 401 error. isAuth will be set to false and deleted from local storage.

FrontEnd/src/hooks/useAuth.js
Lvyshnevska: Done

FrontEnd/src/hooks/useAuth.js
```diff
@@ -1,6 +1,108 @@
...
+
+  useEffect(() => {
+    window.addEventListener('storage', (e) => {
+      if (e.key === 'Token') {
+        setIsAuth(e.newValue);
+      }
+    });
+  });
+
+  const value =  { login, logout, isAuth, isLoading, user, error, mutate };
```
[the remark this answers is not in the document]
Lvyshnevska: Done

FrontEnd/src/hooks/useAuth.js
Lvyshnevska: Do we really have to rename this file? I think we have here both context provider (AuthProvider) and custom hook (useAuth) to provide access to context. Can I leave it unchanged?

FrontEnd/src/tests/CompanyCard.test.js
```diff
@@ -3,6 +3,19 @@ import { MemoryRouter } from 'react-router-dom';
...
 
+const mockedUser = {
+  email: 'test@test.com',
+  id: 1,
+  name: 'Test',
+  profile_id: 1,
+  surname: 'Test'
+};
+jest.mock('../hooks/useAuth', () => ({
+    useAuth: () => {
```
BelousSofiya, line 14: Cool! I like it.

FrontEnd/src/hooks/useAuth.js
popovycholeg: `useAuth` shows us that this file is the hook. But in fact, it is a Context Provider. I would recommend to rename it. There are 3 lines of the `useAuth` hook where we use this context.

popovycholeg: To do not mix white and black we may extract context provider to separate file and leave this hook unchanged

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
```diff
@@ -1,44 +1,27 @@
...
 
     const { data: profileData } = useSWR(
       `${process.env.REACT_APP_BASE_API_URL}/api/profiles/${profileId}?with_contacts=True`,
-      (url) =>
-        fetch(url, {
-          method: 'GET'
+      (url) => axios.get(url)
+        .then(res => res.data)
+        .catch((error) => {
+          console.error('Cannot get profile contact data', error);
```
popovycholeg, line 17: Should we show toast message here?

FrontEnd/src/hooks/useAuth.js
```diff
@@ -1,6 +1,96 @@
...
+    localStorage.removeItem('isAuth');
+    setAuthToken('');
+    delete axios.defaults.headers.common['Authorization'];
+    setIsAuth(false);
+    setUser(null);
+    navigate('/', { replace: true });
+  };
+
+  useEffect(() => {
+    axios.interceptors.response.use(
```
popovycholeg, line 53: 👍

popovycholeg (approved):

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
Lvyshnevska: I think we shouldn't. Notification can confuse user. We are trying to retrieve the data when the component renders. If we don't get the data, the user just won't see the button "показати контакти".

FrontEnd/src/hooks/useAuth.js
Lvyshnevska: Done

PR #426, opened 2024-02-07
#396 Change the type of property edrpou

Lvyshnevska:
The type of EDRPOU field was changed to CharField (as we need to have possibility of adding EDRPOU that starts with zero):
- edrpou_char field was added to the Profile model;
- function for converting existing data of edprou field to str and filling edrpou_char field with it was added to migrations;
- edrpou filed removed, edrpou_char filed renamed to edrpou;
- edrpou validation updated;
- tests updated.


profiles/factories.py
```diff
@@ -1,3 +1,4 @@
+from unittest.util import strclass
```
ProDG, line 1: What is it? Is it used?

profiles/migrations/0008_profile_edrpou_char.py
```diff
@@ -0,0 +1,30 @@
...
+            name="edrpou_char",
+            field=models.CharField(
+                max_length=8,
+                default=None,
+                null=True,
+                unique=True,
+                validators=[validation.validate_edrpou.validate_edrpou],
+            ),
+        ),
+        migrations.RunPython(convert_edrpou_int_filed_to_char, migrations.RunPython.noop),
```
ProDG, line 29: Probably this migration should be elidable?

ProDG (approved):

profiles/models.py
```diff
@@ -77,8 +77,12 @@ class Profile(models.Model):
...
         null=True,
     )
-    edrpou = models.IntegerField(
-        unique=True, validators=[validate_edrpou], default=None, null=True
+    edrpou = models.CharField(
+        max_length=8,
+        unique=True,
+        validators=[validate_edrpou],
+        default=None,
+        null=True,
```
ProDG, line 85: It's usually not a great idea to make nullable field unique. As only one record with null will fit.

profiles/factories.py
Lvyshnevska: Of course it's not used. That's because of autoimport.
Removed it.

profiles/migrations/0008_profile_edrpou_char.py
Lvyshnevska: Done

profiles/models.py
Lvyshnevska: We need to have this field both unique and nullable. As it's not obligatory field during registration and user could fill it later (we have multiple null edrpou fields in db now).

Shushunya (approved):

profiles/models.py
Lvyshnevska: blank=True was added as recommended in documentation for this case

ProDG: @Lvyshnevska wow, if it actually works this way — this is cool. Probably I missed it among PostgreSQL updates, there was a time when it didn't work (you could put only one record with `null` if the field had 'unique' constraint).

PR #439, opened 2024-02-20
#395 Add the possibility to save blank fields

Lvyshnevska:
blank=True and empty string as default value were added to fields that are non mandatory for filling out. This gives the user option to remove previously filled data, if needed, and save the form.


profiles/models.py
```diff
@@ -88,21 +91,23 @@ class Profile(models.Model):
...
         null=True,
+        blank=True,
     )
 
     logo_image = models.ImageField(
         upload_to="logos",
         validators=[validate_image_format, validate_logo_size],
         null=True,
+        blank=True,
     )
```
ProDG, line 111: God help you, guys, with managing `ImageField` (or any other file field) on model which is usually served by API.

Sorry for missing it when it was implemented. Just keep in mind, in general, that if you have a feeling that it was a bad idea — it really was. :)

Better approach: to have a separate model _specifically_ for images (and/or other files). To use a separate endpoints for files upload. And to have FKs to that model for files handling in the core models.

PR #450, opened 2024-03-03
#448 [Authentication] Email validation

Lvyshnevska:
Previously, the email was case-sensitive during registration and login. 
This has been fixed - the email is now validated and stored in db in lowercase during sign-up (validation has been updated), login is carried out with the email in lowercase (TokenCreateSerializer has been overridden for that purpose).
Extra fixes - queryset filter for profiles has been clarified. Only profiles that are not deleted and with active users are returned.

### DIFF
--- a/images/views.py
+++ b/images/views.py
@@ -4,7 +4,7 @@ from rest_framework.generics import (
 from rest_framework.parsers import MultiPartParser, FormParser
 
 from profiles.permissions import UserIsProfileOwnerOrReadOnly
-from profiles.models import Profile, Activity, Category
+from profiles.models import Profile
 from .serializers import BannerSerializer, LogoSerializer
 from utils.completeness_counter import completeness_count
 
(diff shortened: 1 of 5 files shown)


images/views.py
```diff
@@ -13,7 +13,7 @@ class BannerRetrieveUpdate(RetrieveUpdateAPIView):
     permission_classes = (UserIsProfileOwnerOrReadOnly,)
     serializer_class = BannerSerializer
     parser_classes = (MultiPartParser, FormParser)
-    queryset = Profile.objects.all()
+    queryset = Profile.objects.filter(is_deleted=False, person__is_active=True)
```
ProDG, line 16: Such stuff fits nicely into managers (google/investigate, if you're not aware about it).

With managers, you can do stuff like `Profile.objects.only_active().filter(...)` (where `.only_active()` parts applies some common filters).

ProDG (approved):

ProDG (approved):

PR #463, opened 2024-03-08
#378 add the limitation of login

BelousSofiya:
I used library: https://github.com/tomasbasham/ratelimit
For now if user login 10 times in 10 minuts (no matter with wrong or right credentials) this url will be blocked for this user untill 10 minuts will finish.

### DIFF
--- a/authentication/serializers.py
+++ b/authentication/serializers.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 
 from django.contrib.auth import authenticate, get_user_model
 from django.core.exceptions import ValidationError
+from django.conf import settings as custom_settings
 from djoser.conf import settings
 from djoser.serializers import (
     UserCreatePasswordRetypeSerializer,
@@ -19,12 +20,8 @@ from validation.validate_password import (
     validate_password_long,
     validate_password_include_symbols,
 )
-from forum.settings import ATTEMPTS_FOR_LOGIN, DELAY_FOR_LOGIN
 
 User = get_user_model()
-
-# CALLS = ATTEMPTS_FOR_LOGIN
-# DELAY = DELAY_FOR_LOGIN
 
 
 class CustomProfileSerializer(serializers.ModelSerializer):
@@ -106,7 +103,7 @@ class CustomTokenCreateSerializer(TokenCreateSerializer):
         except RateLimitException:
             self.fail("inactive_account")
 
-    @RateLimitDecorator(calls=ATTEMPTS_FOR_LOGIN, period=DELAY_FOR_LOGIN)
+    @RateLimitDecorator(calls=custom_settings.ATTEMPTS_FOR_LOGIN, period=custom_settings.DELAY_FOR_LOGIN)
     def validate_for_rate(self, attrs):
         email = attrs.get(settings.LOGIN_FIELD).lower()
         new_attr = OrderedDict(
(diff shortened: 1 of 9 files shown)


forum/settings.py
```diff
@@ -228,3 +229,8 @@ def running_tests():
     # For tests execution, the fastest password hasher is used
     # (may increase execution speed ×10 and more)
     PASSWORD_HASHERS = ("django.contrib.auth.hashers.MD5PasswordHasher",)
+    DELAY_FOR_LOGIN = 5
+    ATTEMPTS_FOR_LOGIN = 2
+else:
+    DELAY_FOR_LOGIN = 600  # delay time for login in seconds
+    ATTEMPTS_FOR_LOGIN = 10  # attempts for login during delay for login
```
GeorgiyII, line 236: Its OK for now but as for me this is bad practice to have such conditions in settings. If you want to use settings for unittests you can use @override_settings decorator

authentication/serializers.py
```diff
@@ -2,13 +2,18 @@
 
 from django.contrib.auth import authenticate, get_user_model
 from django.core.exceptions import ValidationError
+from django.conf import settings as custom_settings
 from djoser.conf import settings
```
GeorgiyII, line 6: why not `from django.conf import settings as django_settings` and `from djoser.conf import settings as auth_settings` or `from djoser.conf import settings as djoser_settings`

This is much clearer in code when the imports says what about they are.

authentication/serializers.py
```diff
@@ -93,18 +98,18 @@ class Meta(UserSerializer.Meta):
...
-        self.user = authenticate(
-            request=self.context.get("request"), **params, password=password
+    @RateLimitDecorator(
+        calls=custom_settings.ATTEMPTS_FOR_LOGIN,
+        period=custom_settings.DELAY_FOR_LOGIN,
+    )
+    def validate_for_rate(self, attrs):
+        email = attrs.get(settings.LOGIN_FIELD).lower()
+        new_attr = OrderedDict(
+            [("password", attrs.get("password")), ("email", email)]
```
GeorgiyII, line 113: Is there any reasons to use OrderedDict here with two keys?

authentication/serializers.py
BelousSofiya: Thank you for feedback. I've changed names for settings to django_settings and djoser_settings

authentication/serializers.py
BelousSofiya: Thank you for feedback. I've deleted OrderedDict

forum/settings.py
BelousSofiya: This decorator works only if we run single testcase. Smth like this: "python manage.py test authentication.tests.test_user_login". If we run all tests at once (python manage.py test) this decorator would't work. I found this in docs: 
"Warning

The settings file contains some settings that are only consulted during initialization of Django internals. If you change them with override_settings, the setting is changed if you access it via the django.conf.settings module, however, Django’s internals access it differently. Effectively, using override_settings() or modify_settings() with these settings is probably not going to do what you expect it to do."

So, when we are running a specific test case  the override decorator takes effect. When we are running a whole suite specific testcase with override decorator probably isn't the first one that's being run and decorator wouldn't work.
The same for modify_settings() decorator. So I think it's not our varient.

GeorgiyII (approved):

forum/settings.py
BelousSofiya: I'll create a task.

Lvyshnevska (approved):

PR #482, opened 2024-03-26
test setting

BelousSofiya:
Divide settings for dev and tests(to escape if-statement in settings.py file)
New command for running tests:
docker compose -f docker-compose.dev.yml exec api-dev python manage.py test --settings=forum.test_setting

### DIFF
--- a/forum/test_setting.py
+++ b/forum/test_setting.py
@@ -1,0 +1,5 @@
+from .settings import *
+
+PASSWORD_HASHERS = ("django.contrib.auth.hashers.MD5PasswordHasher",)
+DELAY_FOR_LOGIN = 5
+ATTEMPTS_FOR_LOGIN = 2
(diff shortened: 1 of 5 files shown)


forum/test_setting.py
```diff
@@ -0,0 +1,5 @@
+from .settings import *
+
+PASSWORD_HASHERS = ("django.contrib.auth.hashers.MD5PasswordHasher",)
+DELAY_FOR_LOGIN = 5
+ATTEMPTS_FOR_LOGIN = 2
```
ProDG, line 5: Is this the best way to achieve what we need?

It's not very convenient, to use some special settings file to run tests (which we do not only in CI, but also on our local machines).

Frankly, previous implementation looked better for me. Less chances to 'shoot own foot', less chances to have flaky tests which are hard to explain (when user forgets to pass the 'test' settings as a parameter, or just isn't aware about necessity to use some special settings file.

What were the reasons behind making this change?

ProDG (approved):

Lvyshnevska (approved):

PR #492, opened 2024-03-30
#491 Fix the order of companies in the list of saved

Lvyshnevska:
The order of the companies in the list of saved has been corrected.
Saved companies are displayed now in the chronological order (newest first).


profiles/filters.py
BelousSofiya, line 1: It's ok for me, but Igor said that using order_by is not good idea, like Response (for example)

BelousSofiya (approved):

PR #495, opened 2024-03-31
#493 Django Debug Toolbar added to the project

Lvyshnevska:
Django Debug Toolbar has been added to the project.

But I have an issue.
It runs well locally, without containers. 
But when I build all the containers provided by the project, I have to add to the project urls.py:
 
"from django.contrib.staticfiles.urls import staticfiles_urlpatterns
urlpatterns += staticfiles_urlpatterns()"

so that Debug Toolbar and Browsable API work correctly (otherwise static files of drf and debug toolbar are not accessible).

I'm not sure if this is my local problem or if I have to add that to the project code.

### DIFF
--- a/forum/settings.py
+++ b/forum/settings.py
@@ -27,7 +27,7 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 SECRET_KEY = config("SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = config("DEBUG")
 
 ALLOWED_HOSTS = [
     "localhost",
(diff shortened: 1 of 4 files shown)


forum/settings.py
```diff
@@ -37,6 +37,15 @@
...
 ]
 
+if DEBUG:
+    import socket
+
+    hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
+    INTERNAL_IPS = [ip[: ip.rfind(".")] + ".1" for ip in ips] + [
+        "127.0.0.1",
+        "10.0.2.2",
+    ]
```
ProDG, line 47: It looks too complicated and doesn't feel right. You have a 'magic string' here (`10.0.2.2`), plus the rest of the code is not the easiest to read/understand.

Overriding `SHOW_TOOLBAR_CALLBACK` looks like more explicit option. Though, it's a good idea to make it conditional (allowed only when `DEBUG = True` and probably taking some other measures to ensure, that you'll never get DDT turned on at Production environment).

https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#show-toolbar-callback — documentation on the callback options.

At one of my projects, the corresponding part of `settings.py` looks like this:
```python
def show_toolbar(request):
    return not IS_PRODUCTION_ENV and DEBUG and USE_DEBUG_TOOLBAR and not running_tests()


DEBUG_TOOLBAR_CONFIG = {
    'SHOW_TOOLBAR_CALLBACK': show_toolbar,
}
```

ProDG (approved):

PR #500, opened 2024-04-03
serializer, TitleInfo

BelousSofiya:
For long regions: 
image
And for short:
image


Lvyshnevska:
You should also adjust ProfileOwnerDetailViewSerializer to display regions on the user's profile page.
And add style property max-width: 72% to .title-block__company_region to ensure correct size of region block, when many regions are selected.

image

BelousSofiya:
Thank you for feedback. Done

Lvyshnevska (approved):

PR #514, opened 2024-04-09
#215 advanced search

BelousSofiya:
Add functionality of advanced search. It will search companies that contain search word in name, official name, common_info, service_info, product_info fields. '/api/search/advanced?search=t' returns all companies that contain 't' in their fields. Order by name by default.


search/views.py
```diff
@@ -29,3 +29,23 @@ def get_serializer_context(self):
...
+        filters.SearchFilter,
+        filters.OrderingFilter,
+    ]
+    search_fields = [
+        "name",
+        "official_name",
+        "common_info",
+        "service_info",
+        "product_info",
+    ]
```
ProDG, line 49: While the built-in functionality provides relatively wide set of features, in most cases (from my experience — almost always) we're forced to dive into custom implementations anyway. 

So I usually write (kind of) custom filters anyway.

This is NOT an issue, just sharing some facts about the approach I'm using personally.

ProDG (approved):

Lvyshnevska (approved):

parsival2022 (approved):

search/views.py
BelousSofiya: I agree with you, but we don't have enough requirements. Maybe later (after getting requirements from client) we will improve this functionality.

search/views.py
```diff
@@ -29,3 +29,23 @@ def get_serializer_context(self):
             )
             context.update({"saved_companies_pk": saved_companies_pk})
         return context
+
+
```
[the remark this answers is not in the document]
BelousSofiya: Searching is available for all users. Both authenticated or not

Issue #477, opened 2024-03-22
Company/Startup Profile: Add the possibility to copy fields' content


MissMS:
As a user, I want to be able to copy the field value in one click so that I can easily and quickly use it if needed

**Acceptance Criteria**:
1. Add 'Copy' icon before (next to?) the field value in the _'Contact Info'_ block
2. IF the icon Is clicked THEN the value must be copied into the buffer

**Affected fields**: ЄДРПОУ, Телефон, Email, Адреса (**TBD**: is it possible to have multiple addresses?)

**Design:** **to be added**

**Epic link**: #6

PR #521, opened 2024-04-12
#477 add the possibility to copy fields content

parsival2022:
Знімок екрана (184)
Знімок екрана (185)
Знімок екрана (186)
Added icons that, when clicked on, copy phone or email to clipboard. 
Icons provided by Ant Design
When clicked, icon change to green 'clicked' one.
After 4 sec the icon change back to 'copy' one. 
Class .contact-container added to classes.

Path to file with changes: FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx

### DIFF
--- a/FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
+++ b/FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
@@ -57,10 +57,10 @@ function PhoneEmail ({ profileId, personId }) {
                 {isContactsShown || (user && user.id === personId) ? (
                   <div className={classes['data-block__field--contacts']}>
                     <p className={classes['contact-container']}>
-                      <div>{profileData.phone}</div><div onClick={() => copyContent('phone')}>{ isPhoneCopied  ? <CheckOutlined style={{color: '#46a310'}}/> : <CopyOutlined style={{cursor: 'pointer'}}/>}</div>
+                      <span>{profileData.phone}</span><span onClick={() => copyContent('phone')}>{ isPhoneCopied  ? <CheckOutlined style={{color: '#46a310'}}/> : <CopyOutlined style={{cursor: 'pointer'}}/>}</span>
                     </p>
                     <p className={classes['contact-container']}>
-                      <div>{profileData.email}</div> <div onClick={() => copyContent('email')}>{ isEmailCopied  ? <CheckOutlined style={{color: '#46a310'}}/> : <CopyOutlined style={{cursor: 'pointer'}}/> }</div>
+                      <span>{profileData.email}</span> <span onClick={() => copyContent('email')}>{ isEmailCopied  ? <CheckOutlined style={{color: '#46a310'}}/> : <CopyOutlined style={{cursor: 'pointer'}}/> }</span>
                     </p>
                   </div>
                   ) : (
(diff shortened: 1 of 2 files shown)


FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
```diff
@@ -4,11 +4,25 @@ import useSWR from 'swr';
...
+import { CheckOutlined, CopyOutlined} from '@ant-design/icons';
 
 function PhoneEmail ({ profileId, personId }) {
     const [isContactsShown, setContactsShown] = useState(false);
+    const [isPhoneCopied, setIsPhoneCopied] = useState(false);
+    const [isEmailCopied, setIsEmailCopied] = useState(false);
     const { user } = useAuth();
 
+    const copyContent = key => {
+      navigator.clipboard.writeText(profileData[key]);
```
BelousSofiya, line 16: What happens if error?

BelousSofiya (approved):

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
parsival2022: Yes, I agree, the only question is how to handle it. Should we show some modal with error message? I think we need to discuss it and this is related to broader topic of error visual handling.

BelousSofiya: Sometimes we just use 'catch (error) {
      console.error(error);
    }' 
to see error in the console. If later we'll get a desine we'll change it

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
Lvyshnevska, line 1: Shouldn't we display the copy icon only for fields that contain data?

image

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
```diff
@@ -4,61 +4,112 @@ import useSWR from 'swr';
...
+        position: 'top-right',
+        autoClose: 5000,
+        hideProgressBar: false,
+        closeOnClick: true,
+        pauseOnHover: true,
+        draggable: true,
+        progress: undefined,
+        theme: 'light',
+        transition: Bounce,
+      });
```
Lvyshnevska, line 37: You shouldn't add styles for toast messages since we already have global styles for error and success notifications.
Error notifications look like this:
image

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
parsival2022: Agree, it's important detail I missed, will fix that. If no data provided, should we display some placeholder instead?

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
parsival2022: Yes, unnecessery styles will be removed.

FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
Lvyshnevska: No, we shouldn't display placeholders on this page.

Lvyshnevska (approved):

Issue #462, opened 2024-03-06
[Інші Послуги] After clicking on the 'Інші Послуги' tab, it does not navigate to its appropriate section


andymyk:
**Environment:** macOS Ventura 13.5
**Reproducible:** always

**Preconditions**
1. The page http://178.212.110.52/ is open

**Steps to reproduce**
1. Click on the 'Інші Послуги' tab

**Actual result**
Clicking on the 'Інші послуги' tab doesn't navigate to its appropriate section as expected

**Expected result**
Clicking on the 'Інші послуги' tab navigates to its appropriate section as expected

**User story and test case links**
Test Case #461 

<img width="752" alt="2024-03-06_09-19-19" src="https://github.com/ita-social-projects/Forum/assets/87486456/2d3eed41-f039-4790-a956-ebbcfbcb6397">

PR #523, opened 2024-04-12
#462 create a other services tab

ohorodnykostap:
According to bug #462, the operation of the "other services" tab has been corrected.
Also added the generation of the "other services" category to the fixtures.

### DIFF
--- a/FrontEnd/src/components/HeaderFooter/header/menu/HeaderMenu.jsx
+++ b/FrontEnd/src/components/HeaderFooter/header/menu/HeaderMenu.jsx
@@ -20,7 +20,7 @@ const SERVICES_LINKS = [
     },
     {
         title: 'Інші послуги',
-        link: '/profiles/otherServices'
+        link: '/profiles/other-services'
     }
 ];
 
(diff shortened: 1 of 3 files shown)


romanmyko (approved):

FrontEnd/src/components/HeaderFooter/header/menu/HeaderMenu.jsx
```diff
@@ -20,7 +20,7 @@ const SERVICES_LINKS = [
     },
     {
         title: 'Інші послуги',
-        link: '#'
+        link: '/profiles/otherServices'
```
Lvyshnevska, line 23: It is better to avoid joining the words in URLs and use a hyphen to separate them - other-services instead of otherServices.

https://developers.google.com/search/docs/crawling-indexing/url-structure?hl=en&visit_id=638485234540125074-3362765097&rd=1

ohorodnykostap: Thanks for the info.

PR #539, opened 2024-04-23
288 tech debt remove redundant responses in views

parsival2022:
SavedCompaniesCreate and SavedCompanyDestroy views were modified in order to avoid direct hardcoded Response returning. We have choose minimal intervention policy, aiming to use DRF original functionality as much as possible. Therefore only  SavedCompaniesCreate's post() method was customized in order to check if saved company already exists and redirect request to SavedCompanyDestroy view, where the get() method was added in order to recieve redirected request. Also in data request was added user's id before pass it further to create() function.


profiles/views.py
```diff
@@ -52,40 +52,32 @@ class SavedCompaniesCreate(CreateAPIView):
...
+
+        saved_company = SavedCompany.objects.filter(
+            user=user, company_id=pk
+        ).first()
+        if saved_company:
+            return redirect(
+                "profiles:saved_companies_destroy", pk=saved_company.pk
+            )
+        request.data.update({"user": user.id})
+        return self.create(request, *args, **kwargs)
```
ProDG, line 67: Nope.

1. Don't read `request.data` directly (we have serializers for this).
2. Don't manipulate (change) `request.data`. **EVER!**
3. Don't do redirects in REST API.

profiles/views.py
```diff
@@ -52,40 +52,32 @@ class SavedCompaniesCreate(CreateAPIView):
...
-        user = request.user
-        saved_company = get_object_or_404(
-            SavedCompany, company_id=pk, user=user
-        )
-        saved_company.delete()
-        return Response(
-            f"Company {pk} deleted", status=status.HTTP_204_NO_CONTENT
-        )
+    def get(self, request, *args, **kwargs):
+        return super().destroy(request, *args, **kwargs)
```
ProDG, line 80: Please explain, what is going on here.

ProDG (changes requested):

profiles/views.py
parsival2022: We added get() method because we redirecting to this view and redirected request has GET method by default.   This method does exactly the same what delete() do: evokes destroy()

Issue #534, opened 2024-04-18
[FE] Improve fields validation on the Registration form and Edit profile page


Lvyshnevska:
#529 
#544

PR #541, opened 2024-04-23
#534 Improve fields validation on the Registration form and Edit profile page

Lvyshnevska:
Field validation has been updated according to the requirements. 
Notification, that changes cannot be saved due to errors in the form, has been added. 
Submitting changes by pressing Enter in input fields has been disabled on the edit profile page. 

image

image

### DIFF
--- a/FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
+++ b/FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
@@ -146,17 +146,25 @@ const GeneralInfo = (props) => {
     };
 
     const onUpdateField = e => {
-        const fieldValue = e.target.value;
-        const fieldName = e.target.name;
+        const { value: fieldValue, name: fieldName } = e.target;
+        const symbolCount = (fieldValue.replace(/[\s`]/g,'').trim()).length;
         setFormStateErr({ ...formStateErr, [fieldName]: {'error': false, 'message': ''}});
-        if (fieldName === 'name' && fieldValue.length < 2) {
+        if (fieldName === 'name' && symbolCount < 2) {
             setFormStateErr({ ...formStateErr, [fieldName]: {'error': true, 'message': 'Введіть від 2 до 100 символів'}});
         }
-        if (fieldName === 'official_name' && fieldValue.length !== 0 && fieldValue.length < 2) {
+        if (fieldName === 'official_name' && symbolCount !== 0 && symbolCount < 2) {
             setFormStateErr({ ...formStateErr, [fieldName]: {'error': true, 'message': 'Введіть від 2 до 200 символів'}});
         }
         setProfile((prevState) => {
             return { ...prevState, [fieldName]: fieldValue };
+        });
+    };
+
+    const onBlurHandler = (e) => {
+        const { value: rawFieldValue, name: fieldName } = e.target;
+        const fieldValue = rawFieldValue.replace(/\s{2,}/g,' ').trim();
+        setProfile((prevState) => {
+            return { ...prevState, [fieldName]: fieldValue.trim() };
         });
     };
 
@@ -370,6 +378,7 @@ const GeneralInfo = (props) => {
                                 name="name"
                                 label={LABELS.name}
                                 updateHandler={onUpdateField}
+                                onBlur={onBlurHandler}
                                 error={formStateErr['name']?.['error'] ? formStateErr['name']['message'] : null}
                                 requredField={true}
                                 value={profile.name}
@@ -380,6 +389,7 @@ const GeneralInfo = (props) => {
                             name="official_name"
                             label={LABELS.official_name}
                             updateHandler={onUpdateField}
+                            onBlur={onBlurHandler}
                             value={profile.official_name ?? ''}
                             error={formStateErr['official_name']?.['error'] ? formStateErr['official_name']['message'] : null}
                             maxLength={200}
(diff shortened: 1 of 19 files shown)


YanZhylavy (approved):

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -140,8 +146,25 @@ const GeneralInfo = (props) => {
     };
 
     const onUpdateField = e => {
+        const { value: fieldValue, name: fieldName } = e.target;
+        const symbolCount = (fieldValue.replace(/[\s`]/g,'').trim()).length;
```
mehalyna, line 150: Regex replacement `fieldValue.replace(/[\s]/g,'')` removes spaces and backticks. The `trim()` might be redundant because all leading and trailing whitespace would already be removed by the regex

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
```diff
@@ -140,8 +146,25 @@ const GeneralInfo = (props) => {
...
-            return { ...prevState, [e.target.name]: e.target.value };
+            return { ...prevState, [fieldName]: fieldValue };
+        });
+    };
+
+    const onBlurHandler = (e) => {
+        const { value: rawFieldValue, name: fieldName } = e.target;
+        const fieldValue = rawFieldValue.replace(/\s{2,}/g,' ').trim();
+        setProfile((prevState) => {
+            return { ...prevState, [fieldName]: fieldValue.trim() };
```
mehalyna, line 167: The `.trim(`) method is called twice on `fieldValue`. It might be unnecessary

FrontEnd/src/components/ProfilePage/FormComponents/UserInfo.js
```diff
@@ -52,6 +50,34 @@ const UserInfo = (props) => {
...
+        const letterCount = (fieldValue.match(/[a-zA-Zа-щюяьА-ЩЮЯЬїЇіІєЄґҐ]/g) || []).length;
+        const isValidLength = letterCount >= 2 || (fieldName === 'person_position' && letterCount === 0);
+        const isValidPattern = allowedSymbolsPatterns[fieldName].test(fieldValue);
+        let errorMessage = '';
+
+        if (fieldValue && !isValidPattern) {
+            errorMessage = errorMessageTemplates.notAllowedSymbols;
+        } else if (!isValidLength) {
+            errorMessage = errorMessageTemplates.fieldLength;
+        }
```
mehalyna, line 73: If both `!isValidPattern` and `!isValidLength` are true, the latter's message will overwrite the former's, which might not be the intended behavior. Consider handling the case where both conditions could be true and both messages might need to be displayed

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
```diff
@@ -68,12 +73,32 @@ export function SignUpFormContentComponent(props) {
...
+  const validateNameSurname = (value) => {
+    const allowedSymbolsPattern = /^[a-zA-Zа-щюяьА-ЩЮЯЬїЇіІєЄґҐ'\s]+$/;
+    const letterCount = (value.match(/[a-zA-Zа-щюяьА-ЩЮЯЬїЇіІєЄґҐ]/g) || []).length;
+    if (!allowedSymbolsPattern.test(value)) {
+      return errorMessageTemplates.notAllowedSymbols;
+    }
+    if (letterCount < 2) {
+      return errorMessageTemplates.nameSurnameFieldLength;
+    }
+    return true;
```
mehalyna, line 91: The function returns either a string or boolean, which are different types. It would be better to always return the same type for easier handling of the function's return value

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
```diff
@@ -68,12 +73,32 @@ export function SignUpFormContentComponent(props) {
...
+  };
+
+  const onBlurHandler = (fieldName) => {
+    let fieldValue = getValues(fieldName);
```
mehalyna, line 95: It's not clear what happens if `getValues(fieldName)` returns an undefined or null value. It would be good to handle such cases to avoid calling methods on undefined

mehalyna (changes requested):

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
Lvyshnevska: Fixed

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.js
Lvyshnevska: Fixed

FrontEnd/src/components/ProfilePage/FormComponents/UserInfo.js
Lvyshnevska: Updated
image

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
Lvyshnevska: I agree with you, that it would be better to return the same data type. But in this case I might just return an empty string instead of true, and I'm not sure if it is correct. 
If it correct, I'll replace true with an empty string.

ohorodnykostap (approved):

FrontEnd/src/components/SignUp/components/signup/signup-form/SignUpFormContent.js
Lvyshnevska: Check added in if statement.

mehalyna (approved):

Issue #476, opened 2024-03-22
Company/Startup Profile: Update the way how fields are shown on UI


MissMS:
**Problem**: if the length of the email and Address is longer than N symbols the field extends beyond the block size and looks odd
![image](https://github.com/ita-social-projects/Forum/assets/8040919/6c50c06a-5d4f-4ec5-a3e9-2fa61be7565a)

**Acceptance Criteria**:

1. IF '_Email_' is longer than N symbols and it extends beyond the margin of the 'Contact Info' block THEN truncate it, so that it fits the block size and show **<...>** in the end
2. IF '_Address_' is longer than N symbols  and it extends beyond the margin of the 'Contact Info' block THEN split it and span to the next line

Design: **to be added**

**Epic link**: #6 
**Affects**: #61 #295

PR #548, opened 2024-04-24
#476 update the way how fields are shown on UI

KuzykY:
IF 'Email' is longer than N symbols and it extends beyond the margin of the 'Contact Info' and in block 'Update User' THEN truncate it, so that it fits the block size and show <...> in the end
1.1. Add hover on the email field: IF user hovers the field THEN show the full email on the tooltip


FrontEnd/src/components/ProfileDetail/DetailedInfo/PhoneEmail.jsx
```diff
@@ -6,7 +6,9 @@ import { PropTypes } from 'prop-types';
 import classes from './PhoneEmail.module.css';
 import { CheckOutlined, CopyOutlined } from '@ant-design/icons';
 import { toast } from 'react-toastify';
+import {Tooltip} from 'antd';
 
+const LengthEmail = 23;
```
Lvyshnevska, line 11: I think, you should rename this variable to upper case (LENGTH_EMAIL) since it is a constant and add a blank line after its definition.

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/HalfFormField.js
```diff
@@ -1,7 +1,14 @@
-import { PropTypes } from 'prop-types';
+import {PropTypes} from 'prop-types';
 import css from './HalfFormField.module.css';
+import {Tooltip} from 'antd';
+
+const LengthEmail = 32;
```
Lvyshnevska, line 6: The same as above. You should add spaces in the curly braces and rename the constant variable to upper case.

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/HalfFormField.js
```diff
@@ -1,7 +1,14 @@
...
 
 const HalfFormField = (props) => {
+    const shouldShowTooltip = props.name === 'email' && props.value.length > LengthEmail;
+
+    const truncatedEmail = shouldShowTooltip ? `${props.value.slice(0, LengthEmail)}...` : props.value;
```
Lvyshnevska, line 10: If LengthEmail is set to 32 and the email length is longer, there is only one dot after truncated email instead of an ellipsis.
image

Lvyshnevska: I think you should rename variable since you use it for all the fields, like fieldValue.

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/HalfFormField.js
```diff
@@ -13,17 +20,33 @@ const HalfFormField = (props) => {
...
+                        name={props.name}
+                        value={truncatedEmail}
+                        placeholder={props.fieldPlaceholder || 'Введіть текст'}
+                        onChange={props.updateHandler}
+                        required={props.requredField ? 'required' : ''}
+                        disabled={props.name === 'email' ? 'disabled' : ''}
+                        maxLength={props.maxLength}
+                    />
+                )}
             </div>
```
Lvyshnevska, line 50: You can avoid conditional rendering by conditionally setting the tooltip title directly within the property:
<Tooltip title={shouldShowTooltip ? props.value : ''} placement="bottom">

FrontEnd/src/components/ProfilePage/FormComponents/UserInfo.js
Lvyshnevska, line 1: Could you put back all the removed spaces within curly braces, please (imports, object spread).

ohorodnykostap (approved):

VovaD199 (approved):

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/HalfFormField.js
```diff
@@ -15,21 +21,23 @@ const HalfFormField = (props) => {
...
-                    onKeyDown={preventEnterSubmit}
-                    onChange={props.updateHandler}
-                    required={(props.requredField) ? 'required' : ''}
-                    disabled={(props.name === 'email') ? 'disabled' : ''}
-                    maxLength={props.maxLength}
-                />
+                    <Tooltip title={shouldShowTooltip ? props.value : ''} placement="bottom">
+                        <input
+                            type={props.inputType || 'text'}
+                            className={`${css['fields__field--input']} ${props.name === 'email' && css['disabled__field']}`}
```
popovycholeg, line 27: I would recommend using `classnames` package. You might provide an undesired `false` class name.

popovycholeg (approved):

Lvyshnevska (approved):

Issue #294, opened 2023-11-03
Implement User delete functionality


Lvyshnevska:
At the same time as Profile is deleted User should be deleted next way:
is_active - set to False
email - changed to is_deleted_timestamp_email value

PR #566, opened 2024-05-01
#294 Implement user delete functionality

Lvyshnevska:
Simultaneous deletion of profile and user was implemented. 
The max_length property was removed for the email field as we don't have such a requirement.
Minor changes were made on front-end side for correct logout after deleting a profile.


authentication/models.py
```diff
@@ -21,7 +21,7 @@ def create_superuser(self, email, password=None, **extra_fields):
 
 
 class CustomUser(AbstractBaseUser, PermissionsMixin):
-    email = models.EmailField(max_length=50, unique=True)
+    email = models.EmailField(unique=True)
```
ProDG, line 24: Is this correct? Isn't `max_length` mandatory here?

profiles/views.py
```diff
@@ -194,8 +197,14 @@ def get_serializer_class(self):
...
         serializer = self.get_serializer(data=self.request.data)
         serializer.is_valid(raise_exception=True)
-        instance.is_deleted = True
-        instance.save()
+        user = self.request.user
+        with transaction.atomic():
+            instance.is_deleted = True
+            instance.save()
+            user.is_active = False
+            user.email = f"is_deleted_{now()}_{user.email}"
```
ProDG, line 205: I recommend to specify datetime format here. To avoid unnecessary unpredicted issues.

With current implementation, you'll have spaces, `:`, `.`, `+` and God knows what else, + even the length of this can be different for different cases (for example, normally you'll have 6-digits representation of microseconds, but if microseconds were equal to `000000` — they will be omitted in string representation of datetime.

So, `.strftime()` returning timestamp in format `YYYYMMDDhhmmss` would be a cleaner/safer solution.

Also, I'd strongly recommend to have a unittest for this case (checking specific values, by patching `now()` function).

profiles/views.py
```diff
@@ -194,8 +197,14 @@ def get_serializer_class(self):
...
+            user.save()
+            djoser_utils.logout_user(self.request)
```
ProDG, line 207: I don't know anything about Djoser, but — do we need it here? Can't we do just `logout(self.request)`?

ProDG (approved):

authentication/models.py
Lvyshnevska: The email field has a default max_length of 254 characters.

profiles/views.py
Lvyshnevska: Done

BelousSofiya (approved):

profiles/views.py
Lvyshnevska: The authorization token must be removed from db during logout.
logout(self.request) doesn't do it, and djoser_utils.logout_user(self.request) does.

PR #603, opened 2024-05-17
562 [BE] remove redundant responces in views

haodn98:
Remove Responses in Views related to Saved Companies.


profiles/permissions.py
```diff
@@ -29,7 +29,7 @@ def has_permission(self, request, view):
 class IsOwnCompany(BasePermission):
     def has_permission(self, request, view):
         user = request.user
-        pk = request.data.get("company_pk")
+        pk = request.data.get("company")
```
ProDG, line 32: Please create a task to remove all manipulations with request payload and/or query parameters from permission classes.

Permissions must be tied ONLY to the 'core' part of URL and authenticated user.

profiles/serializers.py
```diff
@@ -287,6 +287,25 @@ class SavedCompanySerializer(serializers.ModelSerializer):
...
         model = SavedCompany
         fields = ("id", "user", "company", "added_at")
+        read_only_fields = [
+            "user",
+        ]
+
+    def validate(self, attrs):
+        user = self.context["request"].user
+        company = attrs["company"]
+        company__to_find = SavedCompany.objects.filter(
```
ProDG, line 297: You must not use double underscores in ordinary variable names.

ProDG, line 297: And this variable is redundant, it's used only once in `if`.

ProDG, line 297: And, if you need to check if something exists, you must use `.exists()`.

profiles/tests/test_saved_company.py
```diff
@@ -15,10 +15,9 @@ def test_add_company_to_saved_unauthenticated(self):
         response = self.client.post(
             path="/api/saved-list/",
             data={
-                "company_pk": "{profile_id}".format(
-                    profile_id=self.profile.id
-                ),
+                "company": "{profile_id}".format(profile_id=self.profile.id),
```
ProDG, line 18:
```
"company": self.profile.id
```
?

profiles/tests/test_saved_company.py
```diff
@@ -61,10 +61,11 @@ def test_delete_company_from_saved_authenticated(self):
 
         saved_company = SavedCompanyFactory(user=self.user)
         response = self.client.delete(
-            path="/api/saved-list/{profile_pk}/".format(
-                profile_pk=saved_company.company.id
+            path="/api/saved-list/del/{profile_pk}/".format(
```
ProDG, line 64: You must not use verbs in URLs.

ProDG (changes requested):

profiles/views.py
```diff
@@ -55,41 +54,18 @@ class SavedCompaniesCreate(CreateAPIView):
...
+        return serializer.save(user=self.request.user)
 
 
 class SavedCompaniesDestroy(DestroyAPIView):
     """
     Remove the company from the saved list.
     """
 
+    queryset = SavedCompany.objects.all()
     permission_classes = [IsAuthenticated]
```
YanZhylavy, line 68: Please note that SavedCompaniesDestroy in your implementation performs a lookup by 'pk' (i.e., the subscription id), as it does by default in DestroyAPIView. However, on the frontend, we don't have this information, we need a lookup by the combination of the user id and the company id. If I were you, I would override the get_object() method for SavedCompaniesDestroy.

YanZhylavy (changes requested):

profiles/serializers.py
```diff
@@ -284,9 +284,39 @@ class Meta:
...
+        ]
+
+    def to_internal_value(self, data):
+        company_pk = data.pop("company_pk", None)
+        if company_pk is not None:
+            try:
+                profile = Profile.objects.get(id=company_pk)
+            except Profile.DoesNotExist:
+                raise serializers.ValidationError(
+                    {"non_field_errors": ["Company does not exist"]}
```
ProDG, line 303: Why `"non_field_errors"`?

profiles/tests/test_saved_company.py
```diff
@@ -53,18 +52,22 @@ def test_add_non_existent_company_to_saved_authenticated(self):
...
 
     def test_delete_company_from_saved_authenticated(self):
         self.client.force_authenticate(self.user)
 
         saved_company = SavedCompanyFactory(user=self.user)
         response = self.client.delete(
-            path="/api/saved-list/{profile_pk}/".format(
-                profile_pk=saved_company.company.id
-            ),
+            path=f"/api/saved-list/{saved_company.company.id}/",
```
ProDG, line 68: Revert this change.

profiles/tests/test_saved_company.py
```diff
@@ -53,18 +52,22 @@ def test_add_non_existent_company_to_saved_authenticated(self):
             data={
                 "company_pk": 0,
             },
+            format="json",
         )
         self.assertEqual(status.HTTP_400_BAD_REQUEST, response.status_code)
+        self.assertEqual(
+            {"non_field_errors": ["Company does not exist"]},
```
ProDG, line 59: Why this response is not formatted? 

We have a function `dump()` which allows to get response content easily, do you use it?

profiles/views.py
```diff
@@ -55,40 +54,26 @@ class SavedCompaniesCreate(CreateAPIView):
...
-            f"Company {pk} deleted", status=status.HTTP_204_NO_CONTENT
-        )
+    def get_object(self):
+        queryset = self.get_queryset()
+        lookup_kwargs = {
+            "user_id": self.request.user.id,
+            "company_id": self.kwargs["company_pk"],
+        }
+        obj = get_object_or_404(queryset, **lookup_kwargs)
+        return obj
```
ProDG, line 76: This is bad. DRF allows writing much cleaner (and shorter) code.

You don't have to override `get_object()` just to apply filtering by current user and translation of view kwarg `company_pk` into field name `company_id`.

The first issue is easily addressed by overriding `get_queryset()` (instead of using `queryset` property), the second — by using `lookup_field` and `lookup_url_kwarg` (please read documentation on those).

profiles/tests/test_saved_company.py
```diff
@@ -78,19 +81,25 @@ def test_add_company_to_saved_twice_authenticated(self):
...
             data={
-                "company_pk": "{profile_pk}".format(
-                    profile_pk=self.profile.id
-                ),
+                "company_pk": self.profile.id,
             },
+            format="json",
         )
-        self.client.post(
+        existed_company_response = self.client.post(
```
ProDG, line 88: Why `existed_company_response`?

profiles/serializers.py
```diff
@@ -284,9 +284,45 @@ class Meta:
...
+        ]
+
+    def validate(self, attrs):
+        user = self.context["request"].user
+        company_pk = attrs["company_pk"]
+        if not Profile.objects.filter(pk=company_pk).exists():
+            raise serializers.ValidationError(
+                {
+                    "company_pk": [
+                        f'Invalid pk "{company_pk}" - object does not exist.'
```
ProDG, line 304: I recommend to make validation error messages 'stable' (not having IDs and other variable stuff included). "Company not found" should be OK. There is no need to duplicate information which we have in request.

profiles/serializers.py
```diff
@@ -284,9 +284,45 @@ class Meta:
...
+                    ]
+                }
+            )
+        if SavedCompany.objects.filter(user=user, company=company_pk).exists():
+            raise serializers.ValidationError(
+                {
+                    "non_field_errors": [
```
ProDG, line 311: I believe it's also related to `company_pk` field and is not non-field error.

ProDG (approved):

Lvyshnevska (approved):

YanZhylavy (approved):

PR #611, opened 2024-05-24
Remove redundant responces in views [FE]

YanZhylavy:
Changes in CompanyCard and ProfileCard to communicate with 2 API endpoints - 1 for saving and 1 for deleting saved company.


ohorodnykostap (approved):

romanmyko (approved):

FrontEnd/src/components/CompanyCard/CompanyCard.jsx
```diff
@@ -33,31 +30,23 @@ export default function CompanyCard({
       .map((activity) => activity.name)
       .join(' ');
 
-  async function sendRequest(url, { arg: data }) {
-    return axios.post(url, data);
-  }
-
-  const { trigger } = useSWRMutation(
```
popovycholeg, line 40: Why do we remove swr mutation? We use swr for the state management. Swr mutation updates should update data for `api/saved-lis` key

popovycholeg: After triggering mutation you should update your saved list suing `mutate` function from  `useSWRConfig`

YanZhylavy: I considered using SWR as a state manager and wrote handlers in the following way:

const handleSave = async () => {
  try {
    await axios.post(`${process.env.REACT_APP_BASE_API_URL}/api/saved-list/`, { company: profile.id });
    await mutate(
      (key) => typeof key === 'string' && key.includes('/api/profiles/'),
    );
  } catch (error) {
    console.error(error);
  }
};

const handleDeleteSaved = async () => {
  try {
    await axios.delete(`${process.env.REACT_APP_BASE_API_URL}/api/saved-list/del/${profile.id}`);
    await mutate(
      (key) => typeof key === 'string' && key.includes('/api/profiles/'),
    );
  } catch (error) {
    console.error(error);
  }
};

This was convenient because it updated the state of our target element if it was changed in another browser tab, for example. However, performance was worse due to passing the SWR cache as props to potentially a large number of related elements (if there were many companies on the page). We discussed this with the team during a meeting and decided to stick with a simpler option - using useState. What are your thoughts on this matter? Perhaps there is a better way to write this with SWR?

Issue #655, opened 2024-06-25
Adjust Profile images on the frontend


Lvyshnevska:
Adjust the profile images on the frontend based on the updates made to the backend.
Add "змінити" button next to Banner and Logo fields.
#657

PR #663, opened 2024-07-01
#655 Adjust profile images on the frontend

Lvyshnevska:
Functionality related to images has been adjusted according to backend updates. 
A "змінити" button has been added next to Banner and Logo fields.


FrontEnd/src/components/ProfilePage/FormComponents/FormFields/ImageField.jsx
```diff
@@ -2,62 +2,83 @@ import classNames from 'classnames';
...
-      })}
+  const renderInput = () => (
+    <input
+      accept={accept}
+      id={name}
+      type={inputType}
+      className={css['upload-file__input']}
+      name={name}
+      onChange={updateHandler}
+      onKeyDown={preventEnterSubmit}
```
popovycholeg, line 27: I think prevent default should be added to the submit button. Does the current implementation work correctly with the tab navigation?

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/ImageField.jsx
```diff
@@ -2,62 +2,83 @@ import classNames from 'classnames';
...
+      <img
+        className={css['upload-file__icon']}
+        src={`${process.env.REACT_APP_PUBLIC_URL}/profilepage/camera_icon.png`}
+        alt="Change icon"
+      />
+      <span className={css['update-file__text']}>{text}</span>
+    </label>
+  );
+
+  const renderDeleteButton = (text) => (
```
popovycholeg, line 42: nit: Not sure whether these styles are used anywhere, but it still seems that it's better to use component instead if function.

FrontEnd/src/components/ProfilePage/FormComponents/GeneralInfo.jsx
```diff
@@ -60,8 +60,8 @@ const GeneralInfo = (props) => {
   const { profile: mainProfile, mutate: profileMutate } = useProfile();
   const [profile, setProfile] = useState(props.profile);
   const [formStateErr, setFormStateErr] = useState(ERRORS);
-  const [bannerImage, setBannerImage] = useState(props.profile.banner_image);
-  const [logoImage, setLogoImage] = useState(props.profile.logo_image);
+  const [bannerImage, setBannerImage] = useState(props.profile.banner?.path);
+  const [logoImage, setLogoImage] = useState(props.profile.logo?.path);
   const [bannerImageError, setBannerImageError] = useState(null);
   const [logoImageError, setLogoImageError] = useState(null);
   const [edrpouFieldError, setEdrpouFieldError] = useState(null);
```
popovycholeg, line 67: For complex state logic you may consider using  [useReducer](https://react.dev/reference/react/useReducer)

ohorodnykostap (approved):

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/ImageField.jsx
Lvyshnevska: Yes, current implementation works well for tab navigation. 
Prevent Enter submit was added to all input fields to prevent form being submitted with Enter key, when accidentally pressed in any input field. Because it is unexpected and unwanted button behavior for the user.

FrontEnd/src/components/ProfilePage/FormComponents/FormFields/ImageField.jsx
Lvyshnevska: I suppose such approach is only used in this ImageField component. I'll consider about moving such functions into a separate component.

YanZhylavy (approved):

Issue #666, opened 2024-07-01
Implement Automatic Moderation Time Setting


VovaD199:
Enhance the existing admin page to allow administrators to set the time for automatic moderation by adding a new field "Number of hours for automatic moderation". This field should accept an integer value within a specified range.

**Acceptance Criteria:**

1. **User identification as an Admin**:
1.1. _**Given**_ user has **_is_staff_** characteristic in the DB 
_**When**_ user logs in
_**Then**_ user must be navigated to the Admin screen
AND user must be able to navigate to the regular website from the Admin screen
 
2. **Field Addition:** Add a new field labeled "Кількість годин для автоматичної модерації" to the admin page at http://localhost:8080/customadmin/.
2.1. Default Value: 12 hours
2.2. Valid Value: integer in a range 1 - 48 hours 
   - IF entered value <1, >48, empty THEN show the error message: _'Кількість годин має бути в діапазоні 1-48 години'_

**Documentation Update:**
Update the existing API documentation at http://localhost:8000/api/schema/docs/ to reflect the new field and its constraints.

**Steps to Implement:**

**Database Schema Update:**

Add a new column auto_moderation_hours in the relevant table to store the number of hours for automatic moderation.

**Admin Page Enhancement:**

Update the admin interface at http://localhost:8080/customadmin/ to include the new field.
Ensure the field is labeled "Кількість годин для автоматичної модерації".
Implement validation to restrict input to integers within the specified range.

**Backend Validation:**

Add backend validation to ensure that the value entered in the "Кількість годин для автоматичної модерації" field is an integer between the minimum and maximum values.

**API Documentation:**

Update the API documentation at http://localhost:8000/api/schema/docs/ to include the new field.
Specify the validation rules (integer, minimum, and maximum values) in the documentation.
Testing:

Test the admin page to ensure the new field is displayed correctly and validates input as expected.
Test the automatic moderation feature to ensure it respects the number of hours set by the administrator.

![Image](https://github.com/ita-social-projects/Forum/assets/56256864/d0b191a3-fc9f-4487-8008-ca89ad81d194)

PR #667, opened 2024-07-03
#666 implement automatic moderation time setting

YanZhylavy:
There is only core api stuff for setting automatic moderation delay in this PR, to keep it small. Frontend and Swagger will be updated later


BackEnd/administration/models.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def validate_auto_moderation_hours(value: int):
+    if value < 1 or value > 48:
+        raise ValidationError(
+            "Кількість годин має бути в діапазоні 1-48 години"
+        )
+
+
+class AutoModeration(models.Model):
```
ProDG, line 12: Are there any strong reasons to have such 1-record model, in comparison with just having something like `AUTO_MODERATION_TIMEOUT` in `settings.py`?

ProDG (approved):

Lvyshnevska (approved):

BackEnd/administration/models.py
YanZhylavy: I would do it that way, but we want a moderator who is not a programmer and does not have access to the code to be able to change this value. So we want to create an API endpoint and a corresponding UI interface.

Issue #672, opened 2024-07-07
[Profile Moderation]  Implement Email Notification on Profile Banner/Logo Change


Andrewakiv:
Use Case: Banner/Logo updated
Given Banner and/or Logo is updated
When profile changes are saved
Then approval request must be sent to the Moderator
#616

PR #692, opened 2024-07-13
#672 implement email notification on profile banner/logo change

Andrewakiv:
When the banner or logo is changed, an email is sent to the moderator for review


BackEnd/profiles/templates/profiles/email_template.html
```diff
@@ -0,0 +1,92 @@
...
+            margin-top: 5px;
+        }
+        .banner {
+            width: 100%;
+            max-width: 1500px;
+            height: 200px;
+            background: lightgray;
+            display: flex;
+            justify-content: center;
+            align-items: center;
```
popovycholeg, line 54: Many email apps do not support flexboxes. For the best compatibility for email template development, tables are recommended to align items. Check it https://www.campaignmonitor.com/css/flexbox/align-content/

BackEnd/utils/send_email.py
```diff
@@ -0,0 +1,80 @@
...
+        img.add_header(
+            "Content-Disposition", f'inline; filename="{image_name}"'
+        )
+        email.attach(img)
+
+
+def send_moderation_email(profile):
+    manager = ModerationManager(profile)
+    if manager.check_for_moderation():
+        update_datetime = now() + timedelta(hours=3)
```
GeorgiyII, line 37: If I understand correctly this row responsible for timezone shift, because our timezone is +3 now, am i right?
If yes so when you finish development part and move project to productions so you should spent some time twice a year to change this code line because we have +2 an +3 depends of part of the year. And this is bad idea, our backend should work with own time.

BackEnd/utils/send_email.py
```diff
@@ -0,0 +1,80 @@
...
+        update_time = update_datetime.strftime("%d.%m.%Y %H:%M")
+        update_date = update_datetime.strftime("%d.%m.%Y")
```
GeorgiyII, line 39: I think, this is over-engineering to have different timestamp in context and in email message.
It would. be better to have the same view of this timestamps.

BackEnd/utils/send_email.py
```diff
@@ -0,0 +1,80 @@
...
+        banner = (
+            profile.banner
+            if profile.banner != profile.banner_approved
+            else None
+        )
+        logo = profile.logo if profile.logo != profile.logo_approved else None
+        approve_url = None
+        reject_url = None
```
GeorgiyII, line 47: Why this is urls are None? 
If you use this urls only in 57-58 rows below you don't need to create additional variables.

BackEnd/utils/send_email.py
```diff
@@ -0,0 +1,80 @@
...
+            subject=f"{profile.name} - {update_date}: Запит "
+            "на затвердження змін в обліковому записі компанії",
+            body=email_body,
+            from_email=settings.EMAIL_HOST_USER,
+            to=[
+                settings.EMAIL_HOST_USER,
+            ],
+        )
+
+        email.content_subtype = "html"
```
GeorgiyII, line 72: Such hardcode better to set in beginning of the file like:
EMAIL_CONTENT_SUBTYPE = "html"

BackEnd/utils/image_moderation.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def update_image(self, image, image_type):
+        existing_image = ProfileImage.objects.filter(
+            hash_md5=image.hash_md5, is_approved=True
+        ).first()
+        if existing_image:
+            image.is_approved = True
+            image.save()
+            setattr(self.profile, f"{image_type}_approved", image)
+            completeness_count(self.profile)
```
GeorgiyII, line 19: You will run completeness_count for banner and logo without validation, i think this is not fair, because those images can be declined.

BackEnd/utils/image_moderation.py
```diff
@@ -0,0 +1,32 @@
...
+        self.profile = profile
+        self.moderation_is_needed = False
+
+    def update_image(self, image, image_type):
+        existing_image = ProfileImage.objects.filter(
+            hash_md5=image.hash_md5, is_approved=True
+        ).first()
+        if existing_image:
+            image.is_approved = True
+            image.save()
```
GeorgiyII, line 17: You already filter by is_approved=True, why you set image.is_approved = True again?

BackEnd/utils/send_email.py
```diff
@@ -0,0 +1,80 @@
...
+    if manager.check_for_moderation():
+        update_datetime = now() + timedelta(hours=3)
+        update_time = update_datetime.strftime("%d.%m.%Y %H:%M")
+        update_date = update_datetime.strftime("%d.%m.%Y")
+        banner = (
+            profile.banner
+            if profile.banner != profile.banner_approved
+            else None
+        )
+        logo = profile.logo if profile.logo != profile.logo_approved else None
```
GeorgiyII, line 45: You already checked profile.banner_approved in manager.check_for_moderation()
Maybe it wold be better to update and use that logic?

BackEnd/utils/send_email.py
```diff
@@ -0,0 +1,80 @@
...
+        reject_url = None
+        context = {
+            "profile_name": profile.name,
+            "protocol": "http",
+            "banner": banner,
+            "logo": logo,
+            "updated_at": update_time,
+            "moderation_time": define_ending(
+                AutoModeration.get_auto_moderation_hours().auto_moderation_hours
+            ),
```
GeorgiyII, line 56: I think this logic is over-engineering to set "11 годин" instead of simple python datetime or stripped update_time.
This is bad idea to use localization hard code in code i mean ukrainian language. You already blocked all future abilities for multi language, so will need to refactor huge volume of code.
All localization things should be implemented on UI so your server can give right localized UI based on request parameters or url.

BackEnd/utils/send_email.py
Andrewakiv: We decided to keep the time in UTC.

BackEnd/utils/send_email.py
Andrewakiv: We have decided to keep the time and date separately, but we use the status_updated_at field from the profile.

BackEnd/utils/send_email.py
Andrewakiv: Done, thanks.

BackEnd/utils/send_email.py
Andrewakiv: Done, thanks.

BackEnd/utils/image_moderation.py
Andrewakiv: We have temporarily decided to omit the logic for re-validating images that have already been approved.

BackEnd/utils/image_moderation.py
Andrewakiv: We have temporarily decided to omit the logic for re-validating images that have already been approved.

BackEnd/utils/image_moderation.py
```diff
@@ -0,0 +1,32 @@
...
+            self.profile.status_updated_at = now()
+            self.profile.save()
+            self.moderation_is_needed = True
+
+    def check_for_moderation(self):
+        if self.profile.banner != self.profile.banner_approved:
+            self.update_image(self.profile.banner, "banner")
+        if self.profile.logo != self.profile.logo_approved:
+            self.update_image(self.profile.logo, "logo")
+        return self.moderation_is_needed
```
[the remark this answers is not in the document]
Andrewakiv: Done, thanks.

BackEnd/utils/send_email.py
Andrewakiv: Done, thanks.

BackEnd/utils/send_email.py
Andrewakiv: We will create a separate task to implement this improvement.

Lvyshnevska (approved):

VovaD199 (approved):

YanZhylavy (approved):

Issue #700, opened 2024-07-19
Subscribed startups updates


superolegatron:
As a user, I want to be able to see which startups I am subscribed to have been updated since my last visit. This feature will enhance the user experience by keeping users informed about the latest developments in their subscribed startups. To display updates it should be highlighted or marked on the startups that have been updated.

PR #728, opened 2024-08-15
700 saved companies updates

superolegatron:
implemented saved companies update notification functionality, whenever profile updates, all users who have saved it will be notified. When user has viewed updated company, is_updated sets to False. Updated SavedCompany model, views and serializers for handling corresponding logic. FrontEnd part of tthis task will be in another PR.
#700


YanZhylavy (approved):
Generally, it looks good to me. One minor detail - consider updating the Swagger documentation. Currently, it correctly describes the request body and methods related to subscriptions, thanks to Swagger's automatic analysis. However, it only specifies a single 204 response for both deleting a subscription and patching the 'is_updated' field, because the extend_schema decorator over the view you updated only specifies that response.

Lvyshnevska (approved):

Issue #726, opened 2024-08-14
Handle Moderation request for the removed content


Andrewakiv:
When images under moderation are deleted, an email should be sent indicating that the moderation request has been canceled. If only one image is deleted, the email should be sent with the remaining image still under moderation. Also, update the status to "undefined" when all content is deleted.

PR #729, opened 2024-08-19
#726 handle moderation request for the removed content

Andrewakiv:
When images under moderation are deleted, an email should be sent indicating that the moderation request has been canceled. If only one image is deleted, the email should be sent with the remaining image still under moderation. Also, update the status to "undefined" when all content is deleted.

### DIFF
--- a/BackEnd/utils/moderation/handle_approved_images.py
+++ b/BackEnd/utils/moderation/handle_approved_images.py
@@ -1,0 +1,19 @@
+from ..completeness_counter import completeness_count
+
+
+class ApprovedImages:
+    def __init__(self, profile):
+        self.profile = profile
+
+    def delete_approved_image(self, approved_image):
+        if self.profile.status == self.profile.PENDING and approved_image:
+            approved_image.is_deleted = True
+            approved_image.save()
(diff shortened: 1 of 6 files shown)


BackEnd/utils/moderation/handle_approved_images.py
```diff
@@ -0,0 +1,19 @@
...
+        self.profile = profile
+
+    def delete_approved_image(self, approved_image):
+        if self.profile.status == self.profile.PENDING and approved_image:
+            approved_image.is_deleted = True
+            approved_image.save()
+            completeness_count(self.profile)
+
+    def check_approved_images(self):
+        if self.profile.banner and self.profile.banner.is_deleted:
```
ProDG, line 15: Do I get it right, that if `self.profile.banner.is_deleted == True`, we call `delete_approved_image()` method, which (again) sets `is_deleted = True` and saves the instance?

ProDG (approved):
I'm approving this PR, but method names are pretty confusing, the way the logic is split between classes and methods doesn't look nice.

Lvyshnevska (approved):

PR #781, opened 2024-09-12
768 be incorporate captcha

superolegatron:
Added recaptcha verification from server side, modified signup and sigin serializers


romanmyko:
Add the changes to the READMY.md file. Taking into account the version and the order of use for the developer and devops

Lvyshnevska (approved):

romanmyko (approved):

Issue #822, opened 2024-10-11
View Users List with Sorting Functionality


mehalyna:
### User Story: View Users List with Sorting Functionality

**As an** Admin,  
**I want to** sort the users' list by different columns,  
**So that** I can easily find and manage users based on their name, email, company, status, or registration date.

---

### **Acceptance Criteria:**

1. **API Endpoint for User List with Sorting:**
   - The existing API endpoint for retrieving the users list (e.g., `GET /api/users`) will accept query parameters for sorting.
   - Query parameters:
     - **sort**: Specifies the column to sort by.
     - **direction**: Specifies the sort direction, either `asc` for ascending or `desc` for descending.
   - Example:  
     `GET /api/users?sort=last_name&direction=asc`

2. **Columns Available for Sorting:**
   - The following columns can be sorted:
     - **Прізвище та ім'я (Last Name and First Name):** Sort using the `last_name` field.
     - **Email:** Sort using the `email` field.
     - **Компанія (Company):** Sort using the `company_name` field.
     - **Статус (Status):** Sort using the `status` field.
     - **Дата реєстрації (Registration Date):** Sort using the `registration_date` field.
   
3. **Sorting Mechanism:**
   - Clicking a column header on the frontend triggers a request to the backend with the corresponding sort parameter and direction.
   - If the `direction` parameter is missing, default to **ascending** (`asc`) order.
   - If no sorting parameters are provided, the list is sorted by `registration_date` in descending order (newest first).
   - Example for sorting by status in descending order:  
     `GET /api/users?sort=status&direction=desc`

4. **Database Query Optimization:**
   - Ensure that database queries are optimized by adding indexes on the following columns:
     - `last_name`
     - `email`
     - `company_name`
     - `status`
     - `registration_date`
   - These indexes will ensure efficient sorting even with a large dataset.

5. **Sorting and Filtering Combination:**
   - Sorting can be combined with existing filters (e.g., by status).
   - Example:  
     `GET /api/users?status=active&sort=company_name&direction=asc`  
     This query will return all users with status "active" sorted by `company_name` in ascending order.

6. **Response Format:**
   - The API response includes:
     - `total`: The total number of users matching the filter and sort criteria.
     - `current_page`: The current page of results.
     - `per_page`: The number of users per page.
     - `data`: The list of users sorted according to the specified column and direction.
   - Example JSON response:
     ```json
     {
       "total": 120,
       "current_page": 1,
       "per_page": 20,
       "data": [
         {
[...]

PR #840, opened 2024-10-18
#822 view users list

romanmyko:
Admin panel users list
json format 
```  {
        "id": 47,
        "email": "is_deleted_******_***@test.ua",
        "name": "test",
        "surname": "test",
        "status": {
            "is_active": false,
            "is_staff": false,
            "is_superuser": false,
            "is_deleted": true
        },
        "company_name": "test",
        "registration_date": "2024-10-18"
    }
```
Filter (sort, direction and status)
```
    Filters
    /?id= , /?surname=, /?email= , /?is_active= , /?is_staff=,
    /?is_superuser=,  /?is_deleted=True or False,  /?company_name=, /?registration_date=,
    Ordering sample
    /?ordering=id asc or /?ordering=-id desc
```

### DIFF
--- a/BackEnd/administration/filters.py
+++ b/BackEnd/administration/filters.py
@@ -5,7 +5,7 @@ from django_filters.rest_framework import FilterSet
 class UsersFilter(FilterSet):
     """
     Filters
-    /?id= ,/?name=, /?surname=, /?email= , /?is_active= , /?is_staff=,
+    /?id= , /?surname=, /?email= , /?is_active= , /?is_staff=,
     /?is_superuser=,  /?is_deleted=True or False,  /?company_name=, /?registration_date=,
     Ordering sample
     /?ordering=id asc or /?ordering=-id desc
--- a/BackEnd/administration/filters.py
+++ b/BackEnd/administration/filters.py
@@ -1,6 +1,5 @@
 from django_filters import filters
 from django_filters.rest_framework import FilterSet
-from authentication.models import CustomUser
 
 
 class UsersFilter(FilterSet):
@@ -23,15 +22,3 @@ class UsersFilter(FilterSet):
             ("status", "status"),
         )
     )
-
-    class Meta:
-        model = CustomUser
-        fields = [
-            "id",
-            "name",
-            "surname",
-            "email",
-            "company_name",
-            "status",
-            "registration_date",
-        ]
--- a/BackEnd/administration/filters.py
+++ b/BackEnd/administration/filters.py
@@ -1,7 +1,6 @@
 from django_filters import filters
 from django_filters.rest_framework import FilterSet
 from authentication.models import CustomUser
-from rest_framework import viewsets
 
 
 class UsersFilter(FilterSet):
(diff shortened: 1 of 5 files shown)


BackEnd/administration/filters.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    registration_date = filters.CharFilter(
+        field_name="profile__created_at", lookup_expr="icontains"
+    )
+
+    def filter_is_deleted(self, queryset, name, value):
+        if value:
+            queryset = queryset.filter(email__startswith="is_deleted_")
+        else:
+            queryset = queryset.exclude(email__startswith="is_deleted_")
```
Lvyshnevska, line 30: There is no need for the else statement here.

BackEnd/administration/filters.py
```diff
@@ -0,0 +1,46 @@
...
+            ("id", "id"),
+            ("name", "name"),
+            ("surname", "surname"),
+            ("email", "email"),
+            ("is_active", "is_active"),
+            ("is_staff", "is_staff"),
+            ("is_superuser", "is_superuser"),
+            ("profile__name", "company_name"),
+            ("profile__created_at", "registration_date"),
+            ("email", "is_deleted"),
```
Lvyshnevska, line 44: I guess it should be ("is_deleted", "is_deleted") here.

BackEnd/administration/serializers.py
```diff
@@ -51,15 +51,39 @@ def create(self, validated_data):
...
             "email",
             "name",
             "surname",
+            "status",
+            "company_name",
+            "registration_date",
         )
 
+    def get_company_name(self, obj) -> str:
+        return obj.profile.name if hasattr(obj, "profile") else "No profile"
```
Lvyshnevska, line 71: Wouldn't it be better to return None instead of "No profile" if the user doesn't have a profile?

BackEnd/administration/serializers.py
```diff
@@ -51,15 +51,39 @@ def create(self, validated_data):
...
+
+    def get_registration_date(self, obj) -> str:
+        return (
+            obj.profile.created_at if hasattr(obj, "profile") else "No profile"
```
Lvyshnevska, line 75: The same as above.

BackEnd/administration/views.py
```diff
@@ -29,16 +29,47 @@
 from profiles.models import Profile
 from .permissions import IsStaffUser, IsStaffUserOrReadOnly, IsSuperUser
 
+from django_filters.rest_framework import DjangoFilterBackend as filters
```
Lvyshnevska, line 32: Is there a reason for alias?

BackEnd/administration/views.py
```diff
@@ -29,16 +29,47 @@
...
+            elif status == "superuser":
+                queryset = queryset.filter(is_superuser=True)
+            elif status == "deleted":
+                queryset = queryset.filter(email__startswith="is_deleted_")
+
+        sort = self.request.query_params.get("sort", "name")
+        direction = self.request.query_params.get("direction", "asc")
+        if direction == "desc":
+            sort = f"-{sort}"
+        return queryset.order_by(sort)
```
Lvyshnevska, line 72: There is no need for this method. Filters (django_filters) do all the work.

BackEnd/administration/filters.py
romanmyko: Done

BackEnd/administration/filters.py
romanmyko: Done

BackEnd/administration/serializers.py
romanmyko: Done

BackEnd/administration/serializers.py
romanmyko: Done

BackEnd/administration/views.py
romanmyko: Done

BackEnd/administration/views.py
romanmyko: Done

superolegatron (approved):
Seems ok

Lvyshnevska (approved):

